PR #1, opened 2015-07-10
Current Extrapolation routine

Amy-Xu:
In this PR, I include one link to the dropbox folder that has all SAS and fortran codes, as well as the results excel spreadsheet.


MattHJensen:
Merging.

PR #4, opened 2015-08-17
[DO NOT MERGE] comments for Amy

talumbau:
@Amy-Xu I thought this was a reasonable way to provide comments on the code.

### DIFF
--- a/Stage I/Stage I.py
+++ b/Stage I/Stage I.py
@@ -1,7 +1,12 @@
+# Recommend not having a space in the name of this file
 
 # coding: utf-8
 
+# This is a non-standard thing to do. Often you see 'df' as the generic name
+# for a DataFrame, not as the name for the DataFrame class. The standard thing
+# would be 'from pandas import DataFrame'
 from pandas import DataFrame as df
+
 import pandas as pd
 
 
@@ -12,6 +17,7 @@ import pandas as pd
 # - [Historical estimates from 2000 to 2010](http://www.census.gov/popest/data/intercensal/national/nat2010.html)
 
 #projection 2014+
+# Here you would usually see pd.read_csv, but this looks like it works just as well
 pop_projection = df.from_csv("NP2014_D1.csv", index_col='year')
 pop_projection = pop_projection[(pop_projection.sex == 0) & (pop_projection.race == 0)
                                 & (pop_projection.origin == 0)]
@@ -23,9 +29,16 @@ pop_projection = pop_projection.drop(pop_projection.index[:1], axis=0)
 #estimates 2010-2014
 historical1 = pd.read_csv("NC-EST2014-AGESEX-RES.csv")
 historical1 = historical1[historical1.SEX == 0]
+# in the following line and throughout, the standard way to pass keyword args is with
+# no space between the equal sign, e.g. 'axis=1'. Assignment statements typically
+# leave space
 historical1 = historical1.drop(['SEX', 'CENSUS2010POP', 'ESTIMATESBASE2010'],axis = 1)
 
+# These ages (19, 65, 999) could be extracted to constants at the top
+# of the file. Prevents the 'magic number' phenomenon where you are
+# reading code and just come across various magic numbers
 pop_dep1 = historical1[historical1.AGE<=19].sum()
+# sometimes use see people use the 'inplace=True' flag on drop. Just a heads up
 pop_dep1 = pop_dep1.drop(['AGE'],axis = 0)
 
 pop_snr1 = historical1[(historical1.AGE>=65)&(historical1.AGE<999)].sum()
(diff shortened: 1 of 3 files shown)


talumbau:
After you have a chance to look at (and maybe discuss) this feedback, I could take a crack at putting this in a package, but first I would want to get your thoughts on how you would want to run it.

Stage I/Stage I.py
```diff
@@ -40,16 +53,25 @@
...
 
 
 #combine the estimates of 08-14 with the projection data
+# the 'df' as an indicator that we are making new DataFrames is distracting here
 POP_DEP = pd.concat([df(pop_dep2),df(pop_dep1),df(pop_projection[pop_projection.columns[1:21]].sum(axis = 1))])
 POP_SNR = pd.concat([df(pop_snr2),df(pop_snr1),df(pop_projection[pop_projection.columns[66:]].sum(axis = 1))])
 TOTAL_POP = pd.concat([df(total_pop2), df(total_pop1.values.transpose()),df(pop_projection.total_pop.values)])
 
+#could also do:
+# Stage_II_targets = DataFrame(TOTAL_POP, columns=['TOTAL_POP'])
```
Amy-Xu: That was my first try :'(.....But somehow if I define columns in the same line, all the TOTAL_POP numbers turn into NaN. I have no clue why, and if I separate those into two lines, the problem's automatically fixed. I got stuck here for so long.

screen shot 2015-08-18 at 1 49 36 pm
This is what I got, keeping all other code the same and combining the two lines. Do you know what happened here?

Amy-Xu:
Thank you @talumbau! Your comments make perfect sense to me. I'll change most of them shortly. For the np.where to numba part, do you mean using the vectorize function of numba?

Stage I/Stage I.py
talumbau: This behavior was not obvious to me, but it turns out that when you call the DataFrame constructor with another DataFrame and a `columns` keyword argument, the `columns` argument is interpreted as 'select these columns from the incoming DataFrame' rather than 'impose these names on the new columns'. So, since there is no 'TOTAL_POP' column in the passed in DataFrame, it creates a new empty column with that name. Bad interface, in my opinion.

Stage I/Stage I.py
```diff
@@ -23,9 +29,16 @@
 #estimates 2010-2014
 historical1 = pd.read_csv("NC-EST2014-AGESEX-RES.csv")
 historical1 = historical1[historical1.SEX == 0]
+# in the following line and throughout, the standard way to pass keyword args is with
+# no space between the equal sign, e.g. 'axis=1'. Assignment statements typically
+# leave space
```
Amy-Xu: How about the double equal signs? like 'm==1' or 'm == 1'?

Stage I/Stage I.py
```diff
@@ -40,16 +53,25 @@
 historical2 = historical2.drop(historical2.columns[4:],axis = 1)
 historical2 = historical2.drop(historical2.columns[0],axis = 1)
 
+# Here I would recomment splitting these operations over multiple lines using
+# the &= operator, so something like
+# a = b & c
+# a &= d
+# a &= e
+# instead of a = b & c & d & e
```
Amy-Xu: This made sense to me before, but I got confused when I tried to modify accordingly. For example, pop_dep2 is kind of like a = [b, c], and b = x[m & n].sum(), c = x[h & n].sum(). So it doesn't seem to me the same as a = b & c & e & d &e. Could you explain a bit more?

Stage I/Stage I.py
talumbau: technically for PEP8 compliance, you should have 'm == 1'.

Stage I/Stage I.py
talumbau: Yes, you are right. I read the lines incorrectly. I would try something like this:

```
year1under19 = (historical2.YEAR ==2008) & (historical2.AGE <=19)
year2under19 = (historical2.YEAR ==2009) & (historical2.AGE <=19)
pop_dep2 = []
pop_dep2.append(historical2.TOT_POP[year1under19].sum())
pop_dep2.append(historical2.TOT_POP[year2under19].sum())
```

Stage I/Stage I.py
Amy-Xu: So should I add the column name TOTAL_POP for TOTAL_POP before creating the dataframe, if possible?

Stage II/Single_Year_Stage_II.py
```diff
@@ -13,6 +13,10 @@ def Single_Year_Stage_II(puf, Stage_I_factors, Stage_II_targets, year, tol):
 
     print("Preparing coefficient matrix...")
 
+    # the first half of this function (with all of the np.where statements)
+    # might be amenable to refactoring to a numba function, could improve
+    # readability
```
Amy-Xu: Just want to confirm, here you mean the @vectorize decorator in numba package?

talumbau: Oh I did not think of `@vectorize`, although that may be a good idea. I meant that it may be possible to use @jit or perhaps even our custom @iterate_jit (or a similar technique) to expand the np.wheres to to be 'if' statements and so on.

Amy-Xu: Can I give `@vectorize` a try first? I feel if all inputs have the same dimension, it might work. If it doesn't work out, I'll let you know asap. And then we can switch to `@jit`. How does that sound to you?

talumbau: I think that's a good idea, especially if everything has the same dimensionality.

martinholmer:
Closing #9 which is not really a pull request.

PR #5, opened 2015-08-19
[WIP]Modifications according to TJ's comments

Amy-Xu:


Stage II/Single_Year_Stage_II.py
```diff
@@ -128,8 +124,13 @@ def Single_Year_Stage_II(puf, Stage_I_factors, Stage_II_targets, year, tol):
...
-            WAGE_1,WAGE_2, WAGE_3,WAGE_4, WAGE_5, WAGE_6, WAGE_7,WAGE_8,WAGE_9, WAGE_10, WAGE_11, WAGE_12]
+    temp = [Single, Joint, HH, SS_return, Dep_return,
+            INTEREST,DIVIDEND, BIZ_INCOME, BIZ_LOSS, CAP_GAIN, ANNUITY_PENSION,
+            SCH_E_INCOME, SCH_E_LOSS, SS_INCOME, UNEMPLOYMENT_COMP,
+            WAGE_1,WAGE_2, WAGE_3,WAGE_4, WAGE_5, WAGE_6,
+            WAGE_7,WAGE_8,WAGE_9, WAGE_10, WAGE_11, WAGE_12]
+    
+    b = []
     for m in temp:
         b.append(m)
```
talumbau: I still wasn't 100% sure on the difference between `temp` and `b`, but if you want a new list `b` that has the same contents as the list `temp`, I would do `b = list(temp)`. A minor change, but just something I noticed.

talumbau:
this looks good to me, aside from my one small comment above

Stage II/Single_Year_Stage_II.py
Amy-Xu: Whoops, I forgot to change to change this line. Originally temp and b are different on the first five variables (Single, Joint, HH etc). I added those variables to temp in this PR so temp and b are exactly the same, so that I can replace the for loop. Somehow I forgot the most important part :P

Amy-Xu:
@talumbau I have a quick question. 

The example in vectorize documentation looks like this:
`from numba import vectorize, float64`
`@vectorize([float64(float64, float64)])`
`def f(x, y):`
     ` return x + y`

If I have more than one array to return, how should I add more float64 to the signature? I didn't find much documentation on this part.

talumbau:
Sorry I left you hanging here. One suggestion I would have would be to have some of the input arguments also be output arguments. This is sort of a C-style way of doing things, but it is frequently done with lower level functions. The idea is that your actual function might return void (so nothing returned) but your function may take 6 arguments, for example, with three of them being input arguments and 3 of them being so-called "output arguments". Let me look around quickly for an example of this with numba.

talumbau:
hmm. Ok I looked into this a bit. Basically the answer is that the multiple outputs would need to be "stacked" into a single 2D array output, which is somewhat inconvenient. So, that would push me towards suggesting `@jit` but you may not be able to compile it with `nopython` mode w/o some refactoring of the function. It might not be worth it depending on how much you care about the performance of this section. You could always just try `@jit` by itself and see if there is enough performance improvement just with that alone.

talumbau:
to be clear, you would need to use the `@guvectorize` decorator and make your last argument a m-by-n array where `m` is the length of the passed in vectors and `n` is the number of outputs you are returning. Some documentation here:

http://numba.pydata.org/numba-doc/dev/reference/compilation.html

Amy-Xu:
@talumbau It seems this is more complicated than I thought. Switching to @guvectorize is probably going to take me a while. As we just got 09 PUF yesterday and need to do the extrapolation for this file soon, do you think we could leave this part for now and merge in the other commits? In terms of the performance, it takes about half a minute for each year after 08, including this np.where section and the LP solving process. It seems to be fine for now. May be we can come back to optimize it later?

talumbau:
sounds very good to me!

Amy-Xu:
Thank you @talumbau. Will open another issue for this after the extrapolation of 09 PUF. Merging.

PR #11, opened 2015-12-21
Added script for fixing recid 

Amy-Xu:
This PR created a new folder, final prep, for all the steps after Stage II. Though many other processing will be added, currently this folder only includes one script that makes RECID unique for the CPS matched file. Please refer to the comments of this script for more details on implementation.

@martinholmer Do you think this is better? Could you review?

### DIFF
--- a/final prep/puf-cps-processing.py
+++ b/final prep/puf-cps-processing.py
@@ -1,0 +1,47 @@
+# The recid (key) for each record is not unique after CPS match
+# because 1) original PUF records were split in some cases
+# 2) non-filers were added
+
+# This script fixes the recid issue by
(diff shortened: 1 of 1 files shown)


martinholmer:
@Amy-Xu said:
> @martinholmer Do you think this is better? Could you review?

Script ```puf-cps-processing.py``` is just the same as before, right?
If so, then I think everything looks good for a merge into the master branch.

I would do that, but I don't seem to have commit privileges on this repo.
Why don't you do the commit now?  That way, we can move on to the next steps.

Amy-Xu:
@martinholmer yep they're the same. I guess I'm the only person who can merge now. :P Merging.

PR #13, opened 2016-03-01
Updated script 1) remove extra variables 2) adjust variable names for 08 PUF

Amy-Xu:
@martinholmer Could you please review the script when you get a chance? Thanks

### DIFF
--- a/final prep/puf-cps-processing.py
+++ b/final prep/puf-cps-processing.py
@@ -17,6 +17,7 @@ This script transforms the raw csv file in several ways as described below.
 import argparse
 import sys
 import pandas
+import numpy
 
 
 def main():
@@ -133,7 +134,7 @@ def remove_unused_variables(data):
                         'e11900', 'e18600', 'e25960', 'e15100', 'p27895', 'e12200'}
     data = data.drop(UNUSED_READ_VARS, 1)
 
-    data = data.replace('            ', 0)
+    data = data.fillna(value=0)
     return data
 
 def transform_variables_to_09(data):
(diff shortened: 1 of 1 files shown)


martinholmer:
Amy,

This looks good, but I haven't been able to test the files it produces.
Did you want to handle the eight-thousand-some missing variables as part
of this pull request?

Best wishes, Martin

Amy-Xu:
Earlier this week Matt told me there's going to be a updated 09 file, and theoretically there shouldn't be any missing values anymore. I'm thinking probably we can wait a bit on that part?

martinholmer:
OK.  That sounds fine.

Amy-Xu:
@martinholmer Feedback from SOI says those values simply should be zeros. I think it's also how we're dealing with them now, right?

martinholmer:
No. They are BLANK, which is not the same as zero in Pandas.  So, could you make the missing values zero in the final prep script in the tax data repo?

Amy-Xu:
@martinholmer Can I replace them manually? Somehow neither regex nor string could get the blanks replaced........

martinholmer:
Amy,
I'm confused.  I thought you said that Matt told you that John O'Hare was going to fix this problem soon.

But if you want to fix this problem in the ```puf-cps-processing.py``` script, then won't something like this work (warning: I haven't tested this code fragment)?
```
df[df[varname]==np.nan] = 0.
```

Won't that assign zero for every NaN (missing value) for the dataframe column with ```varname```

cc @Amy-Xu @MattHJensen

martinholmer:
@Amy-Xu,  Why don't you go ahead and merge taxdata pull request #13 into the taxdata master branch?

Amy-Xu:
Thanks for reminding @martinholmer. 

Merging.

PR #36, opened 2016-09-13
Remove four more unused variables from puf.csv file

martinholmer:
This taxdata-repository pull request removes four additional variables from the `puf.csv` input file used by Tax-Calculator and TaxBrain.  These four variables are no longer used in Tax-Calculator logic, and therefore, are completely redundant.  A Tax-Calculator-repository pull request will be prepared that matches this one by dropping these four variables from the `Records.VALID_READ_VARS` set.  Once that pull request is merged, only the smaller `puf.csv` file generated by this pull request will work with Tax-Calculator.

@MattHJensen @feenberg @talumbau @Amy-Xu @GoFroggyRun @andersonfrailey @codykallen

PR #41, opened 2016-10-26
Add randomly assigned fourth dependent under 13 and elderly dependent

andersonfrailey:
This PR along with PR #36 are being used to update the PUF file used by TaxCalc. It randomly assigns a fourth under 13 dependent to households with three dependents, and randomly assigns an elderly dependent to some households. These variables are needed specifically in [Tax-Calulator PR #999](https://github.com/open-source-economics/Tax-Calculator/pull/999).

The decimals used for the random assignment come from the 2014 CPS. 

When assigning the fourth dependent, I primarily relied on the variable `hunder15`, which gives the number of persons under 15 in each household because it was the closest I could get to 13. I first found the number of households that had at least three people under 15 then found what fraction of that number had at four or more under 15 (about one third). I then looked at the number of children under 13 that lived in households with three or more people under 15 and found that slightly more than one third of that number were in households with four or more under 15. The number I have used for the random assignment is the average of these two.

Similarly, for the elderly dependent assignment I found what fraction of those over 65 in the CPS were listed as dependents.

### DIFF
--- a/final prep/puf-cps-processing.py
+++ b/final prep/puf-cps-processing.py
@@ -337,7 +337,16 @@ def add_dependents(data):
     Randomly assign fourth under 13 dependent to households with three or
     more dependents under 13 and randomly assign an elderly dependent
 
-    Decimals used when randomly assigning dependents come from 2014 CPS data
+    Decimals used when randomly assigning dependents come from 2014 CPS data.
+    Using the 'hunder15' variable (the closest one available to under13),
+    roughly 1/3 of the number of households with three or more persons under
+    15 had four or more under 15. Looking at those ages 13 and under in the
+    data set, slightly over 1/3 of those in households with three or more
+    under 15 were in households with four or more under 15. The decimal used
+    here is the average of these two numbers.
+
+    Similarly, the number used in elderly dependent assignment is the fraction
+    of those 65 and older that are listed as dependents in the data set.
     """
     # Randomly assign fourth dependent to units with at lease three dependents
     np.random.seed(409)
--- a/final prep/puf-cps-processing.py
+++ b/final prep/puf-cps-processing.py
@@ -75,7 +75,10 @@ def main():
     data['e02100p'] = earnings_split * data['e02100']
     data['e02100s'] = one_minus_earnings_split * data['e02100']
 
-    # (E) Remove variables not expected by Tax-Calculator
+    # (E) Randomly assign additional dependents to households
+    data = add_dependents(data)
+
+    # (F) Remove variables not expected by Tax-Calculator
     if max_flpdyr >= 2009:
         data = remove_unused_variables(data)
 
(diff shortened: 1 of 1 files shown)


andersonfrailey:
@Amy-Xu could you take a look at this and merge if you believe it's ready?

Amy-Xu:
The code looks good to me. Do you mind adding your description in this PR as a part of comments in your function? So our users won't need to trace all the way to this PR to see where you get the two hard-coded numbers.

andersonfrailey:
@Amy-Xu taken care of.

Amy-Xu:
LGTM. Merging. Thanks Anderson!

martinholmer:
Well,  I guess I'm too late (got the new `puf.csv` just an hour ago), but for the record I'd like to point out what seems to be a problem: there are more under-13 kids in some filing units than there are kids eligible for the child tax credit (which means under 17, right?).  This newest version of the `puf.csv` does fix the excessive number of filing units with 4+ under-13 kids.  But it seems, now that the imputation code has been made available, there are filing units with 3+ under-13 kids who have only three child credit kids who are then imputed to have 4+ under-13 kids.  Seems like we should have skipped those for the imputation.  But maybe this is OK because we cannot (given the IRS-SOI censoring of the PUF N24 data) really know how many are age-eligible for the child tax credit.  Here are the details:

What IRS-SOI says about the N24 variable:
```
Counts have been limited depending on filing status. 
For joint and head of household returns the limit is 3,
for single returns the limit is 2 and for married filing
separately returns the limit is 1.  These limits are 
applied sequentially by type of dependent, starting
with children living at home, then children living away
from home, then other dependents, then parents.
```
And here are tabulations of the newest `puf.csv` file:
```
$ awk -F, 'NR==1{for(i=1;i<=NF;i++)print i,$i}' puf-2016-10-28.csv | grep nu13
81 nu13
$ awk -F, 'NR==1{for(i=1;i<=NF;i++)print i,$i}' puf-2016-10-28.csv | grep n24
10 n24
$ awk -F, 'NR==1{for(i=1;i<=NF;i++)print i,$i}' puf-2016-10-28.csv | grep s006
71 s006
$ awk -F, 'NR>1{n[$81]+=$71}END{for(i in n)print i,n[i]}' puf-2016-10-28.csv
0 1.298e+10
1 1.64961e+09
2 9.58285e+08
3 2.01205e+08
4 1.15812e+08
$ awk -F, 'NR>1&&$81>$10{n++}END{print n}' puf-2016-10-28.csv
1445
$ awk -F, 'NR>1&&$81>$10{t++;n[$81]++}END{print t;for(i in n)print i,n[i]}' puf-2016-10-28.csv
1445
1 166
2 113
3 28
4 1138
```
So, 1138 out of 1445 filing units with `nu13 > n24` have been imputed to have 4+ under-13 kids.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @GoFroggyRun

PR #50, opened 2016-12-06
Fix ss_income calculation

andersonfrailey:
This PR adds a filter for filers to `Single_Year_Stage_II.py` to properly calculate ss_income.

@Amy-Xu

### DIFF
--- a/Stage II/Single_Year_Stage_II.py
+++ b/Stage II/Single_Year_Stage_II.py
@@ -33,7 +33,7 @@ def Single_Year_Stage_II(puf, Stage_I_factors, Stage_II_targets, year, tol):
     annuity_pension = puf.e01700*s006
     sch_e_income = np.where(puf.e02000>0, puf.e02000, 0)*s006
     sch_e_loss = np.where(puf.e02000<0, -puf.e02000, 0)*s006
-    ss_income = puf.e02400*s006
+    ss_income = np.where(puf.filer == 1, puf.e02400, 0)*s006
     unemployment_comp = puf.e02300*s006
 
 


martinholmer:
@andersonfrailey,  Could you explain this change to someone who is not on the inside of the data preparation activity, but who will be a user of the prepared data?  In particular, the proposed change in PR#50 seems to imply (although I could be totally confused) that we are going to ignore social security benefits received by those who are not income tax filers.  Is that correct? If that is correct, why are we doing that?  I'm quite sure that non-filers from the CPS do have social security benefits received as part of their survey information.

@Amy-Xu @MattHJensen @feenberg

Amy-Xu:
@martinholmer The script here is to adjust wage and salary distribution, and at the same time to make sure AGI items are not threw off because of the adjustments. The target for social security income (e02400) is from SOI, which only provides the estimate for filers. We currently don't have any target for non-filer social security income and let their SS grow at CBO projection rate without any adjustment.

Amy-Xu:
@martinholmer Vaguely speaking, every user of TC is a user of the prepared data here because this script produces the [weights file](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/WEIGHTS.csv) required by TC.

martinholmer:
@Amy-Xu said:
> The script here is to adjust wage and salary distribution, and at the same time to make sure AGI items are not thrown off because of the adjustments. The target for social security income (e02400) is from SOI, which only provides the estimate for filers. We currently don't have any target for non-filer social security income and let their SS grow at CBO projection rate without any adjustment.

OK, I see what you're doing, **but** SSA must have aggregate totals for OASDI benefits paid in each calendar year.  Why not use that extra information to make the prepared data more accurate?

@andersonfrailey @MattHJensen

Amy-Xu:
@martinholmer Martin, I think we definitely should add this target at some point as non-filers social security income is quite huge. But right now everyone's packed and non-filer's social security doesn't seem dragging off any tax liabilities. We might want to start on this as soon as we clear the top priority list.

martinholmer:
@Amy-Xu said about taxdata pull request #50:
> Martin, I think we definitely should add this target at some point as non-filers social security income is quite huge. But right now everyone's packed and non-filer's social security doesn't seem dragging off any tax liabilities. We might want to start on this as soon as we clear the top priority list.

OK.

Amy-Xu:
Thanks @andersonfrailey @martinholmer!

PR #54, opened 2016-12-16
Fill nan values with zeros in CPS file

andersonfrailey:
This PR adds one line of code to `cps-processing.py` that will fill `NaN` values with zeros. Without this, any calculations involving a variable that has `NaN` values will result in `NaN`. 
@Amy-Xu


Amy-Xu:
Great! Thanks Anderson!

PR #59, opened 2017-02-03
[WAIT] Rename directories and .py files

andersonfrailey:
This PR follows up a phone call I had with @martinholmer yesterday. It renames the TaxData directories and python files to replace spaces with underscores and move away from using roman numerals in directory and file names. 

@martinholmer, should I rename the .csv files used in stage one and stage two with roman numerals also or is that unnecessary?

cc: @Amy-Xu @MattHJensen


martinholmer:
@andersonfrailey, This (PR #59) is much better, but it seems as if with the coming of the cps.csv file (and its own associated Stage1 blowup factors, Stage2 sample weights, and Stage3 adjustment ratios), we need a directory structure something like this:
```
cps_data          <---- currently called "CPS File Prep"
cps_stage1
cps_stage2
cps_stage3
cps_finalprep
cps_docs
puf_stage1        <---- currently called "Stage I"
puf_stage2        <---- currently called "Stage II"
puf_stage3
puf_finalprep     <---- currently called "final prep"
puf_docs          <---- currently called "docs"    
```
Does something like this make sense?

@MattHJensen @Amy-Xu

martinholmer:
@andersonfrailey,  Let's wait on the renaming of directories and files until we get your several pending pull requests merged into the master branch.

@MattHJensen

andersonfrailey:
@martinholmer said:
>Let's wait on the renaming of directories and files until we get your several pending pull requests merged into the master branch.

Sounds good.

As for restructuring the directory for the CPS, I can't say with any certainty yet, but I'd imagine there will be some overlap in the files used on `puf.csv` and `cps.csv`, so that will need to be taken into consideration. Other than that looks good to me.

martinholmer:
@andersonfrailey said:
> As for restructuring the directory for the CPS, I can't say with any certainty yet, but I'd imagine there will be some overlap in the files used on puf.csv and cps.csv, so that will need to be taken into consideration. Other than that looks good to me.

Yes, I can see what you mean.  For example, the CSV files containing population and economic targets will be the same most likely.  Could we put them in their own directory?  What would we call it?  Something like `target_data`?  Can you think of a better name?  And then that `target_data` directory could also contain any documentation about the source of the population target data (from Census, etc.) and about the source of the economic target data (from CBO projections and other places).

martinholmer:
@andersonfrailey,  I've thinking more about your comment on #59.  Here is my current view about the new directory structure:
```
target_data
cps_data          <---- currently called "CPS File Prep"
cps_stage1
cps_stage2
cps_stage3
cps_finalprep
cps_docs
puf_stage1        <---- currently called "Stage I"
puf_stage2        <---- currently called "Stage II"
puf_stage3
puf_finalprep     <---- currently called "final prep"
puf_docs          <---- currently called "docs"    
```
Even though this plan has cps_docs and puf_docs directories, it seems to me we should have documentation files in the same directory as the data or script is located.  So, for example, any documentation about the nature and source of target data should be in the target_data directory.

martinholmer:
Pull request #59 has been superseded by the merger of #62 and #64.

PR #73, opened 2017-02-12
Add to script/input/output documentation for README.md files

martinholmer:
Adds some documentation, but no README.md links to documentation files.


martinholmer:
@andersonfrailey,  Can you follow-up merged taxdata pull request #73 by moving the files in the taxdata/docs/info directory into the appropriate doc subdirectory.  See, for example, the new location of your markdown documentation of PUF stage3, which has been moved to the `taxdata/puf_stage3/doc/puf_stage3.md` file.  Note how the `taxdata/puf_stage3/README.md` file has been edited to point to each file in the doc subdirectory.

Please consider clarifying the file names as you move them to their new doc subdirectory.  And also, consider improving the formatting of the text files (e.g., make line length be be less than 80 characters as in PEP8) and consider clarifying the substance of the text.

If any files in taxdata/docs/info directory do not have an obvious destination to which to move them, please raise an issue at the taxdata repo.

And finally, there are two files in the taxdata/docs/notebook directory; can they be moved into a doc subdirectory?  Our goal is to move all the file in the taxdata/docs directory, so that directory can be removed.

@MattHJensen

andersonfrailey:
@martinholmer, I'm finishing up moving all the documents and fixing the formatting of the text files. The notebooks can be moved from `taxdata/docs/notebooks`, but there are three documents in taxdata/docs/info that don't necessarily fit in just one directory.

`Extrapolation_description.pdf`, `ExtrapolationMethodology.pdf`, and `MatchingDocumentationRevised.pdf` all describe both Stage 1 and 2 of the extrapolation process so it wouldn't make sense to include them in just on subdirectory as in. I've talked to @MattHJensen about rewriting these documents in a non-pdf format (likely as .md or LaTeX files) so when doing that the sections could be separated and each file added to the corresponding subdirectory. Until that time I would suggest we either keep `taxdata/docs` or have them sit outside any subdirectory with the `README.md` file for the entire repo. Thoughts?

martinholmer:
@andersonfrailey said:
> `Extrapolation_description.pdf`, `ExtrapolationMethodology.pdf`, and `MatchingDocumentationRevised.pdf` all describe both Stage 1 and 2 of the extrapolation process so it wouldn't make sense to include them in just on subdirectory as in. I've talked to @MattHJensen about rewriting these documents in a non-pdf format (likely as .md or LaTeX files) so when doing that the sections could be separated and each file added to the corresponding subdirectory. Until that time I would suggest we either keep taxdata/docs or have them sit outside any subdirectory with the README.md file for the entire repo. Thoughts?

I'll let you and Matt decide on how high a priority to assign rewriting these three documents.  Strikes me that this is pretty low priority relative to generating the new `cps.csv` file and its associated files, adding more variables to Stage 3 adjustments, updating `puf.csv` to a year more recent than 2009, etc.

Meanwhile, let those three PDF files remain in the taxdata/docs/info directory and I'll try to figure out some sensible way to handle them.  

Just for clarification, the `MatchingDocumentationRevised.pdf` is about matching CPS records (to represent non-filers) to the IRS-SOI PUF and has nothing to do with Stage 1 or Stage 2, right?  Or am I missing something?

andersonfrailey:
@martinholmer 
>Just for clarification, the MatchingDocumentationRevised.pdf is about matching CPS records (to represent non-filers) to the IRS-SOI PUF and has nothing to do with Stage 1 or Stage 2, right? Or am I missing something?

This is correct. When we finish translating the matching program from SAS to python this will be rewritten and can be moved to the repo holding that code.

PR #76, opened 2017-02-13
Movedocs

andersonfrailey:
This PR moves documentation from `taxdata/docs/info` to the subdirectories each file is associated with, fixes the formatting of text documents, deletes the unnecessary `Data_Sources` Excel file, and moves notebooks from `taxdata/docs/notebooks` to their associated subdirectories.

At this time, there are no documents address strictly the stage 2 process, which is why there is no docs folder added to `puf_stage2`. As I mentioned in a comment in PR #73, the documents that describe the stage 2 methodology also describe stage 1. When these are rewritten the sections can be split so that documentation can be added.

The comments on the last commit are also misleading. The program I use automatically fills the commit message with the last one you used and I just didn't change it for this one.

@martinholmer


martinholmer:
@andersonfrailey,  Thanks for the quick work on the taxdata documentation reorganization.
I'll merge these changes now.

PR #82, opened 2017-03-04
Remove APOPN and AGDPN from final growfactors.csv file

martinholmer:
This taxdata pull request will be merged after [Tax-Calculator pull request 1214](https://github.com/open-source-economics/Tax-Calculator/pull/1214) is merged.


martinholmer:
QUESTION: who has the document that was used to generate the [taxdata/doc/Stage1_Stage2_2009PUF.pdf](https://github.com/open-source-economics/taxdata/blob/master/doc/Stage1_Stage2_2009PUF.pdf)?

That document needs two sets of changes:
- In Appendix A, the AGDPN factor should be deleted.
- In Appendix B, the line with `AGDPN  e03240` needs to be deleted and a new line in the `ATXPY` section with `e03240` needs to be added.

If somebody can send me the document, I'll be happy to make the changes.

@MattHJensen @Amy-Xu @andersonfrailey

andersonfrailey:
> QUESTION: who has the file that was used to generate the Extrapolation Overview PDF document?
@martinholmer, I believe @Amy-Xu has it.

martinholmer:
@andersonfrailey said:
>  I believe @Amy-Xu has it [the file that was used to generate the Extrapolation Overview PDF document]?

Thanks for the info.

@Amy-Xu, do you want to make the documentation changes described in taxdata #82?  Or, if you send me the file, I'll be happy to make the changes myself and pass the file on to @andersonfrailey for safe keeping.  How would you like to proceed?

@MattHJensen

Amy-Xu:
Just handed the original document to Martin. But I think it might worth a bit time thinking for this type of pdf documents. Should we keep the original editable version together somewhere? Also this file originally was created in a rtf editor, as we thought rtf could be displayed directly on github. But the math functions didn't turn out to be right. In the end still keep the semi-mark-down-rtf format in the end and then convert it to a PDF. Should we actually turn the original version into word, as it's more accessible than rtf editors? Although there're free rtf editors online: http://rst.ninjs.org

Issue #92, opened 2017-06-08
Python Translation Initial Results


andersonfrailey:
I just finished running the python-produced PUF file through tax-calc. [Here](http://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/New%20PUF%20Analysis.ipynb) is a notebook containing the results. Is there anything else people would like added to the notebook?

Most of the results are almost exactly the same. There is a slight dip in the number of returns and tax units with zero/negative income and combined tax liability. The latter is connected to the former because if you look at the number of units with zero/negative tax liability as a percent of the total returns they are similar. The drop in total number tax-units is also almost entirely non-filers, the sum of which is cut roughly in half.

The top decile also sees a noticeable increase in tax liability. I haven't pinned down the root cause of this yet. If you look at the distribution of tax liability it's actually almost exactly the same across the board.

There's one change I need to make to the scripts that adds a variable containing the number of dependents under 18 to address [Tax-Calc issue #1409](https://github.com/open-source-economics/Tax-Calculator/issues/1409). After that, I'll open a PR to merge everything in when I'm finished making changes to `puf_data/finalprep.py` and `puf_stage2/stage2.py`.

@MattHJensen @martinholmer @Amy-Xu @hdoupe

martinholmer:
@andersonfrailey, Please coordinate with me before you issue a new Python-generated version of the `puf.csv` file because I want to add some code to the `final_prep.py` script that hopefully will eliminate the need in Tax-Calculator for the out-of-place `ID_Casualty_frt` data adjustment.

@MattHJensen

martinholmer:
@andersonfrailey,  Whey are there only **half** the number of non-filers in the Python-generated file relative to the old SAS-generated file?
```
filers_py = (calc_py.records.filer * calc_py.records.s006).sum()
filers_sas = (calc_sas.records.filer * calc_sas.records.s006).sum()
nonfilers_py = (calc_py.records.s006[calc_py.records.filer == 0]).sum()
nonfilers_sas = (calc_sas.records.s006[calc_sas.records.filer == 0]).sum()
print '# Filers, Python: {}'.format(filers_py)
print '# Filers, SAS:    {}'.format(filers_sas)
print '# Non-Filers, Python:  {}'.format(nonfilers_py)
print '# Non-Filers, SAS:    {}'.format(nonfilers_sas)
# Filers, Python: 153581159.34
# Filers, SAS:    153581143.76
[...]

PR #93, opened 2017-06-12
Add num dep under 18

hdoupe:
This feature estimates the number of dependents under the age of 18 in order to address [Tax-Calculator issue #1409](https://github.com/open-source-economics/Tax-Calculator/issues/1409).

### DIFF
--- a/puf_data/finalprep.py
+++ b/puf_data/finalprep.py
@@ -393,12 +393,12 @@ def add_dependents(data):
 
     # Count number of dependents under 18
     def reassign(agedp):
-        # With probability 0.46 remove members from age group
+        # With probability 0.47 remove members from age group
         # 17 and 18. This probability is the proportion
         # of 18 year olds in this age group.
         if agedp != 4:
             return agedp
-        return 1 if np.random.rand() <= 0.460867 else 0
+        return 0 if np.random.rand() <= 0.465920 else 4
 
     agedp1 = data.agedp1.apply(reassign)
     under1 = np.where(((agedp1 > 0) & (agedp1 <= 4)), 1, 0)
@@ -409,12 +409,15 @@ def add_dependents(data):
     agedp3 = data.agedp3.apply(reassign)
     under3 = np.where(((agedp3 > 0) & (agedp3 <= 4)), 1, 0)
 
-    # randomly add dependents
-    # the probability decreases significantly after 5
-    choice = [3, 4, 5]
-    prob = [0.91, 0.084, 0.006]
+    # Randomly add dependents according to distribution
+    # observed in CPS
+    # The proportion of tax units with at least three
+    # dependents under age 18 with 0, 1, or 2 more dependents
+    # under age 18
+    choice = [0, 1, 2]
+    prob = [0.910, 0.0828, 0.00633]
     add_rand = np.where(under1 == 1 & under2 == 1 & under3 == 1,
-                        np.random.choice(choice, p=prob) - 3, 0)
+                        np.random.choice(choice, p=prob), 0)
 
     nu18_dep = under1 + under2 + under3 + add_rand
     data["nu18_dep"] = nu18_dep
--- a/puf_data/finalprep.py
+++ b/puf_data/finalprep.py
@@ -391,7 +391,43 @@ def add_dependents(data):
     under5 = age1 + age2 + age3
     data['nu05'] = under5
 
-    # Count number of dependents under 18
+    return data
+
+
+def add_nu18_dep(data):
+    """
+        This function estimates the number of dependents under
+        the age of 18 for a given tax unit. This relies on
+        the agedp* variables and is done in two stages.
+            1. Classify dependents as under 18 years old
+            based off of the value of agedp* and random selection
+            from the 17-18 age group.
+            If agedp* is 1, 2, or 3, then this dependent
+            is in age range 0-16.
+            If agdep* is 4, then this dependent is either 17
+            or 18 years old. To estimate the number of 17 year
+            olds in this age group, agedp* is set to 0 (and
+            thus is no longer considered in nu18_dep count)
+            with probability 0.47 which is the proportion
+            of 18 year olds in this age group as observed in
+            the CPS.
+            If agedp* is greater than 4, then this dependent is
+            over age 18 and is not considered in the nu18_dep count.
+
+            2. Estimate the number of dependents under the age
+            of 18 in the case where there are three
+            known dependents under the age of 18. The PUF file
+            tells us the age of up to three dependents. In the CPS,
+            of tax units with at least three dependents under
+            the age of 18, 91% only have three dependents,
+            8% have 4 dependents, and 0.6% have 5 dependents
+            under the age of 18 (the proportion decreases
+            substantially after 5). According to these
+            proportions either 0, 1, or 2 is added to the number
+            of known dependents.  If there are less than three
+            dependents under the age of 18, then no dependents are
+            added.
+    """
     def reassign(agedp):
         # With probability 0.47 remove members from age group
         # 17 and 18. This probability is the proportion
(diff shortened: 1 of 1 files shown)


martinholmer:
@hdoupe,  Even though I've read over [Tax-Calculator issue #1409](https://github.com/open-source-economics/Tax-Calculator/issues/1409), I'm still having difficultly following what you are doing in the taxdata code you're proposing to add in pull request #93.  Can you add some comments to your pull request that explain what the new code does?  

Also, why name the new variable `nu18_dep`?  Why not just `nu18`?  Isn't that the way the other analogous variables are named?

@MattHJensen @andersonfrailey

martinholmer:
@hdoupe,  I strongly suggest that you put your proposed code and comment in a new function.

@MattHJensen @andersonfrailey

hdoupe:
@martinholmer Thank you for your input.  I moved my code into a new function and added comments to better explain what the code does.  

`nu18` is already a variable and it counts the number of individuals under 18 years old in a tax unit.  I used `nu18_dep` to emphasize that this variable  counts the number of dependents under the age of 18.

martinholmer:
@hdoupe said:
> `nu18` is already a variable and it counts the number of individuals under 18 years old in a tax unit. I used `nu18_dep` to emphasize that this variable counts the number of dependents under the age of 18.

Oh.  How many of the `nu18` are not in `nu18_dep`?
I wasn't aware that we made such a distinction.
Why would there be non-dependent members of a filing unit?
Doesn't the term "filing unit" strongly suggest all members of the filing unit are dependents?

@MattHJensen @feenberg @Amy-Xu @andersonfrailey

hdoupe:
@martinholmer, for 97% of tax units these variables are the same.  But there are some cases where they differ.  Just looking through the code, there could be an independent filer who is under 18 or a spouse under 18 who isn't a dependent.  I treated `nu18_dep` the same way that `nu05` and `nu13` are treated.  

Here is the breakdown: 
image

@MattHJensen @feenberg @Amy-Xu @andersonfrailey

martinholmer:
@hdoupe said:
> for 97% of tax units these variables are the same. But there are some cases where they differ. Just looking through the code, there could be an independent filer who is under 18 or a spouse under 18 who isn't a dependent. I treated `nu18_dep` the same way that `nu05` and `nu13` are treated.

Thanks for the quick tabulation and explanation.

But I'm still puzzled.  What does it mean to be `an independent filer who is under 18 or a spouse under 18 who isn't a dependent`?  Isn't an "independent filer" in his/her own filing unit?
How can a spouse not be a dependent?  If they are married filing separately, the spouse will be in his/her own filing unit.  Sorry to be so dense about this but it doesn't make sense to me.

And what is the weighted count for the three percent?  Are these filing units with relatively large or relatively small sampling weights?

@MattHJensen @feenberg @Amy-Xu @andersonfrailey

hdoupe:
@martinholmer, you're right. It is strange.  I'll let someone with more knowledge in this area weigh in here.

Here's the breakdown after applying the weights:
image

And the code to produce this:
`df[nu18_diff] = df.nu18 - df.nu18_dep

def wt_count():

    vc = df.nu18_diff.value_counts()
    value = vc.index.tolist()
    count = vc.values.tolist()

    wted = []

    for v, c in zip(value, count):
        wts = df.loc[df.nu18_diff == v, 'wt']
        wted.append(wts.sum()*c)

    wted = np.array(wted)
    wted = wted/wted.sum()

    return pd.DataFrame.from_dict({'values': value, 'wted': wted})


if __name__ == '__main__':
    res = wt_count()
`

MattHJensen:
@martinholmer and @hdoupe, 

My understanding is that the primary taxpayer and spouse are not considered dependents of their own tax filing unit. (although they might be a dependent of another tax unit). The layout of 1040 supports this:
image


Here is more info: https://www.irs.gov/publications/p17/ch03.html
image

martinholmer:
@MattHJensen said:
> My understanding is that the primary taxpayer and spouse are not considered dependents of their own tax filing unit. (Although they might be a dependent of another tax unit.) The layout of 1040 supports this.

OK, what's the implication of this for the work @hdoupe is doing in taxdata pull request #93?

Seems like we already have an pretty accurate `nu18_dep` in the existing `nu18` variable.
So, it seems to me we can dispense with #93 and use `nu18` without giving up much accuracy.

This approach seems particularly promising because the new Python-generated `cpi-matched-puf.csv` file will not need the logic in #93 according to @andersonfrailey in #92.  So, why add it now and have to drop it in a week or two?

@feenberg @Amy-Xu

MattHJensen:
> OK, what's the implication of this for the work @hdoupe is doing in taxdata pull request #93?

 I was answering your question, "But I'm still puzzled. What does it mean to be an independent filer who is under 18 or a spouse under 18 who isn't a dependent? Isn't an "independent filer" in his/her own filing unit? How can a spouse not be a dependent?" The implication is that `nu18` is conceptually different from `nu18_dep.`

> Seems like we already have an pretty accurate nu18_dep in the existing nu18 variable.
So, it seems to me we can dispense with #93 and use nu18 without giving up much accuracy.

That makes sense to me given the closeness of `nu18_dep` and `nu18`. @andersonfrailey, are you comfortable with including the `nu18` variable in a  new SAS-generated puf.csv file.

martinholmer:
@MattHJensen said:
> I was answering your question, "But I'm still puzzled. What does it mean to be an independent filer who is under 18 or a spouse under 18 who isn't a dependent? Isn't an "independent filer" in his/her own filing unit? How can a spouse not be a dependent?" 
**The implication is that nu18 is conceptually different from nu18_dep.**

So, you're saying in a married filing jointly filing unit where the spouse is less than 18 years old, the spouse will be included in the `nu18` count but should not be included in the `nu18_dep` count.
Is that correct?  Is that a sensible construction of the `nu18` variable?

martinholmer:
**More about the definition of `nu18`**

The Tax-Calculator `records_variables.json` file contains this:
```
    "nu05": {
      "type": "float",
      "desc": "Number of dependents under 5 years old",
      "form": {"2013-2016": "imputed from CPS data"}
    },
    "nu13": {
      "type": "int",
      "desc": "Number of dependents under 13 years old",
      "form": {"2013-2016": "imputed from CPS data"}
    },
    "nu18": {
      "type": "int",
      "desc": "Number of people under 18 years old in the filing unit",
      "form": {"2013-2016": "imputed from CPS data"}
    },
    "n1821": {
      "type": "int",
      "desc": "Number of people over 18 and under 21 years old in the filing unit",
      "form": {"2013-2016": "imputed from CPS data"}
    },
    "n21": {
      "type": "int",
      "desc": "Number of people 21 years old or older in the filing unit",
      "form": {"2013-2016": "imputed from CPS data"}
    },
```
Given that `nu05` and `nu13` are counts of **dependents**, why isn't `nu18` also a count of dependents?   I guess I forgot why the `nu18`, `n1821` and `n21` variables switched from dependents to people (including young spouses apparently).  Why the difference in definition?

@MattHJensen @Amy-Xu @andersonfrailey @hdoupe

hdoupe:
>    Seems like we already have an pretty accurate nu18_dep in the existing nu18 variable.
    So, it seems to me we can dispense with #93 and use nu18 without giving up much accuracy.

>>That makes sense to me given the closeness of nu18_dep and nu18. @andersonfrailey, are you comfortable with including the nu18 variable in a new SAS-generated puf.csv file.

This makes sense to me, too. In hindsight, I should have checked this before adding the variable in taxdata.  Thanks for reviewing.

@martinholmer @MattHJensen

martinholmer:
@hdoupe said:
> This makes sense to me, too.  In hindsight, I should have checked this before adding the variable in taxdata.  Thanks for reviewing.

Do we need to add any Tax-Calculator code to implement the request in [Tax-Calculator issue 1409](https://github.com/open-source-economics/Tax-Calculator/issues/1409) using `nu18`?

hdoupe:
@martinholmer I think that we just have to subtract `nu18` from `XTOT` when the personal exemption is calculated in the AGI function in `functions.py`.

martinholmer:
Close taxdata pull request #93 since using `nu18` to approximate `nu18_dep`.

martinholmer:
Closing #93

martinholmer:
Close taxdata pull request #93 since using `nu18` to approximate `nu18_dep`.

Issue #106, opened 2017-06-15
Extrapolate welfare data


Amy-Xu:
The third task [outlined in this issue](https://github.com/open-source-economics/UBI-examples/issues/9) is to develop an extrapolation routine for welfare data in CPS tax unit dataset. An initial thought is to assume for each program, participation and benefit grow at respectively X and Y percent each year, where X and Y are derived from historical data. (If official projection targets are available, then we could use those targets directly.) Then could use the same logit regression for imputation to meet the targets for participation growth and then apply an uniform ratio to everyone in order to blow up total benefit. 

Many details need to be considered, but for now the most tricky part is whether to do this extrapolation on tax unit or original program benefit unit (individual/household) in raw CPS. Individual or household level is natural since all projection or historical data would be available at these level; however, this will create enormous difficulty afterwards because raw CPS needs to go through tax-unit creation process. The weights of records do not stay the same over time and thus extrapolation based on 2014 raw CPS weight cannot guarantee hitting the targets in later years. On the other hand, extrapolating the data at tax-unit level would make later steps easier, but there isn't any targets or historical welfare data at tax-unit level. 

Other things to consider:
- Should we consider population growth the same way as tax-data extrapolation?
- Non-filer targets? Since many benefit programs would require non-filer numbers to be accurate.

Any thoughts? @MattHJensen @martinholmer @andersonfrailey @hdoupe

Amy-Xu:
Have been collecting [targets & historical data](https://docs.google.com/spreadsheets/d/1COz47zqm-rSUn5kM7QSJtHk6CmcUI5Uv--bxs8wCu20/edit?usp=sharing), and will keep updating this spreadsheet.

Amy-Xu:
When clarifying this extrapolation problem to @hdoupe last Friday, I realized there might be a way to do the extrapolation at 'tax unit' level without participation targets at tax unit level. I thought it through again and haven't seen any loophole yet. So would love to hear how everyone feels about this plan:

1. Attach individual level probability (derived from logit regression based on 2014 data) that indicates program participation to CPS dataset in the tax-unit creating process, and in this way obtain each tax-unit member's likelihood of participation in one program or another.

2. Add or remove participation according to the probability attached, to meet the participation targets assumably derived from the official projection or historical data (see the spreadsheet from the comment above). Since this imputation is based on the weights of all tax units in 10 years, we need no adjustment afterward.

3. Then adjust the benefit total according to the targets.

[...]

PR #108, opened 2017-07-11
extrapolate CPS benefits

hdoupe:
This script extrapolates participation and benefits for welfare programs as discussed in issue #106.

This PR is a WIP and still needs to be augmented so that it can handle other types of benefits besides SSI.

cc @Amy-Xu


Amy-Xu:
I think it worth a bit time to see where is the best place for this extrapolation routine script. @martinholmer @andersonfrailey 

But that's  not a high priority at the moment.

andersonfrailey:
Bringing this back up, I think the best option is to keep this extrapolation routine in a `cps_stage3` folder. This way we can run everything in `cps_data` to create the raw CPS tax unit file, then create the weights using code in the `cps_stage2` folder, then run benefit extrapolation, then clean up the CPS file so that it's ready for Tax-Calculator.

andersonfrailey:
I'm also working on extending the extrapolation to include 2027 now that we have that capability.

andersonfrailey:
>I'm also working on extending the extrapolation to include 2027 now that we have that capability.

Scratch that. We can only extrapolate to 2027 with the PUF file, not the CPS. Mental lapse on my part

martinholmer:
@andersonfrailey said:
> We can only extrapolate to 2027 with the PUF file, not the CPS. 

Can you explain why that is the case?  Does the benefit extrapolation only go thru 2026?

If so, is this going to cause a problem in Tax-Calculator?  
How is Tax-Calculator going to handle input data sets with two different last years?

@MattHJensen @Amy-Xu

andersonfrailey:
@martinholmer asked:
>Can you explain why that is the case? Does the benefit extrapolation only go thru 2026?

This is because the weights file we have for the CPS only goes to 2026. John gave us the code he used to create the file, but not the code to create the weights file so I can't generate a new one that goes to 2027 on my own.

As part of PR #118, I'm working on a script to create an accompanying weights file, but it isn't complete yet.

I haven't given too much thought to how this would affect Tax-Calculator. I'd imagine if you were to try and run Tax-Calculator to 2027 using the CPS you would get an error that the specified column in the weights file doesn't exist.

I will also email John to see if he has a new weights file available to goes to 2027.

andersonfrailey:
Now that the merge conflict has been resolved, I'm planning on merging this PR tomorrow unless there are additional comments.

Amy-Xu:
Thanks Anderson!

PR #111, opened 2017-07-20
Add SAS files to make CPS

andersonfrailey:
This PR contains all of the SAS scripts used to create the CPS dataset. Instructions for how to use them are included in the README file. Unfortunately the directories aren't named in accordance to standard practice, but all of the file paths in the script themselves are coded to work with it so I didn't want to cause any issues by renaming everything right now.

An unfortunate side effect of merging this is the repo may be classified as a SAS repo again.

@MattHJensen @martinholmer


martinholmer:
@andersonfrailey,  Thanks for all the work.  I was able to generate the exact same two cps*csv.gz files as included in this pull request #110.  Merging into the master branch.

Issue #104, opened 2017-06-14
Alternatives to CyLp


andersonfrailey:
Before PR #100 was merged there was some discussion of finding an alternative to CyLp to use in stage 2 of our extrapolation process. @hdoupe and I have talked about this offline as well. I'm opening this issue so we have a place to discuss other open source options we may have.

I did a quick search after lunch, and if there is no opposition to moving away from Python, there is another CLP interface written in Julia, another open source language. I haven't looked too deeply yet, but judging from the [repository](https://github.com/JuliaOpt/Clp.jl), it is a relatively immature project, but a pull request was opened just 13 days ago so there is more activity going on than in CyLp.

I don't have much experience with Julia, but I'm working through the getting started manual and it's relatively easy to pick up the syntax and was built specifically for numerical computing so it's very fast.

I will keep looking for alternatives in my spare time and if others find any I would appreciate them posting them here for discussion. I'm sure there are more options that the one I've posted, possibly in R or another computational language.

@martinholmer @Amy-Xu

martinholmer:
@andersonfrailey and @hdoupe,
Thanks for following up on this which-linear-programming-library-to-use issue.

My main frustration was not being able to install `cylp`, but now I know about easy_install and all is good.  We should probably add that bit of information in the `taxdata/README.md` file.  This is the only "extra" Python package needed to work with the taxdata repo, right?

Never underestimate the value of a package that actually **works**.  There are many packages that are only striving to work, so we should change away from `cylp` only after careful consideration.

@MattHJensen @feenberg @Amy-Xu

hdoupe:
I created a new [branch](https://github.com/hdoupe/taxdata/tree/linprog_experiments) to facilitate testing alternatives to CyLp.  I separated the `solve_lp_for_year` function into a function that prepares `A1`, `A2`, and `b` and a function that takes `A1`, `A2`, `b`, and `tol` and solves for `r` and `s`.

I also looked at writing a solver in Julia (`solve_lp_for_year.jl`).  This doesn't work yet, but if anyone is interested in using Julia for this problem, then they could start from this file.  The IO works, and I wrote what I thought might work for the [JuMP](https://jump.readthedocs.io/en/latest/) solver. 

@andersonfrailey @martinholmer @Amy-Xu

andersonfrailey:
I believe we should reopen discussion on replacing CyLp. I tried to run `stage2.py` yesterday and received this error:
```
Traceback (most recent call last):
[...]

PR #113, opened 2017-08-16
[WIP] New Linear Programming Model

andersonfrailey:
Following up on the discussion in issue #104, I started testing out a new linear programming model for stage 2 of the extrapolation process. This PR is a first pass at using the [PuLP](https://pythonhosted.org/PuLP/index.html) package.

I've marked it [WIP] because I'm still looking into ways to speed up calculations. I ran it today and it took roughly 5 hours to finish. Other than that, it seems to work relatively well. I put together a notebook to compare calculations when using the current weights file and when using the weights created with PuLP. That can be seen [here](http://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/LP%20Comparisons%20-%20PULP.ipynb).

I believe the main addition slowing everything down is this loop:

```python
for i in range(len(b)):
    LP += pulp.lpSum([(A1[i][j] * r[j] + A2[i][j] * s[j])
                     for j in puf.index]) == b[i]
```

Now that I have a slightly better feel for the package I have a couple of ideas on how to avoid needing that loop. I'll post updates here as I work.

Also, it appears Scipy just merged the PR with a new linear programming model so it should be available in their next release. When that happens I'll also test out their model.

cc @hdoupe @martinholmer @Amy-Xu

### DIFF
--- a/puf_stage2/solve_lp_for_year.py
+++ b/puf_stage2/solve_lp_for_year.py
@@ -1,11 +1,8 @@
 import numpy as np
-from cylp.cy import CyClpSimplex
-from cylp.py.modeling.CyLPModel import CyLPArray, CyLPModel
+import pulp
 
 
 def solve_lp_for_year(puf, Stage_I_factors, Stage_II_targets, year, tol):
-
-    puf_length = len(puf.s006)
 
     print("Preparing coefficient matrix for year {} .....".format(year))
 
@@ -69,8 +66,8 @@ def solve_lp_for_year(puf, Stage_I_factors, Stage_II_targets, year, tol):
                               wage_11, wage_12))
 
     # Coefficients for r and s
-    A1 = np.matrix(One_half_LHS)
-    A2 = np.matrix(-One_half_LHS)
+    A1 = np.array(One_half_LHS)
+    A2 = np.array(-One_half_LHS)
 
     print("Preparing targets for year {} .....".format(year))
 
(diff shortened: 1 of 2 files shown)


martinholmer:
@andersonfrailey,  This is good you're testing alternative LP packages.

So, is it reasonable conclude from your PULP-vs-current-PUF results that lower income people are getting bigger weights so that the number of filing units is up 7.5 million (an increase of 4.6%), income tax revenue is down by 2.7% and payroll tax revenue is up by 7.1% (an increase of $68.3 billion)?  If not, what do you think is going on?

Do you have a sense of how changing LP packages can make such big difference?  Or, at least to me, these seem like "big" changes.

@Amy-Xu @hdoupe

andersonfrailey:
@martinholmer I believe your conclusion may be correct. I looked at the average weight by AGI bin and you seen an uptick in almost all of the bins except for the very lowest and very highest. The average AGI in the bins that seen an increase in average weight ranges from 7,774 to 130,024 and average wages range from 5,669 to 95,833.

You can see the full tables in an updated version of the [notebook](http://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/LP%20Comparisons%20-%20PULP.ipynb).

> Do you have a sense of how changing LP packages can make such big difference? Or, at least to me, these seem like "big" changes.

I do not right now. I'm not familiar enough with the inner workings of the algorithms they use to solve the optimization problems to give a definitive answer to this question.

andersonfrailey:
I'm going to go ahead and close this PR. It's significantly behind the rest of the repo and while I still plan on introducing a new linear programming model soon it will be easier to work off a fresh PR than this one.

PR #118, opened 2017-10-20
CPS creation script 

andersonfrailey:
This PR deletes all of the SAS scripts used to create the CPS file and replaces them with python versions. I've labeled it "WIP" because I'm double checking a few aspects of the script to make sure everything works properly, but I wanted to open it up for review now.

There are a lot of hard coded numbers in the script. This is just a direct translation from the SAS scripts. Next time I talk to John I'm going to try and find out where they all came from.

I didn't include the CPS files in DAT format that are used as inputs because they are rather large files and can be downloaded from NBER. If people would prefer they are added I'm happy to do so. I'm also still looking for ways to speed up the processing of the DAT files. The CPS comes in a non-delimitated  format where each line in the file represents a record in the CPS. The solution I have right now reads each line of the DAT file as a string and pulls individual variables based on their location in the string. Because the files are so large and there are so many variables, it takes about 2 hours for each CPS file. If anyone has any ideas on how to speed this process up, please let me know. In the meantime I've written the scripts so that after the first time you run them it creates a CSV of the CPS files and on all subsequent runs will read those in instead of processing the DAT files.

The basic structure is:
`cpstaxunits`: folder within `cps_data` that contains all of the python scripts used to create the CPS file. Everything is run when you run `create cps.py`. There is one file not used directly in the creation script, `read_state_data.py`. This script reads IRS tax information on returns at the state level to create targets used in the creation process.

`data`: folder within `cpstaxunits`that contains all of the coefficients used during the imputation process and C-TAM generated benefit files. This is also the folder that the CPS files should be saved after being downloaded from NBER.

I haven't created a comparison notebook between the CPS produced by the SAS scripts and this new one yet. I'll do that in the coming days. There also is not an accompanying weights file at this time. I still need to write those scripts.

I've been working on this for awhile now so it's likely I neglected to explain something in this post. If there's anything that needs clarification, please let me know.

@martinholmer @MattHJensen @Amy-Xu


andersonfrailey:
Latest commit adds code to rename variables so that they can be used in the benefit extrapolation script. See #120 for more details.

andersonfrailey:
I forgot I had this branch and started working on the CPS creation scripts separately so this branch is woefully out of date. I'll be deleting this branch and closing this PR and opening a separate one for all future work on this project.

PR #120, opened 2017-11-01
Update final prep scripts to include benefit data

andersonfrailey:
This PR updates the final prep scripts to include benefit data in the final CPS file. It also accounts for variables that needed to be renamed in order to be compatible with the benefit extrapolation scripts @hdoupe and @Amy-Xu wrote. I will be updating my CPS file PR to include those name changes shortly.

### DIFF
--- a/cps_data/finalprep.py
+++ b/cps_data/finalprep.py
@@ -48,12 +48,12 @@ def main():
         'BLIND_HEAD': 'blind_head',
         'BLIND_SPOUSE': 'blind_spouse',
         'HMIE': 'e19200',
-        # 'SSI': 'ssi_ben',
-        # 'VB': 'vet_ben',
-        # 'MEDICARE': 'mcare_ben',
-        # 'MEDICAID': 'mcaid_ben',
-        # 'SS': 'ss_ben',
-        # 'SNAP': 'snap_ben',
+        'SSI': 'ssi_ben',
+        'vb_ben': 'vet_ben',
+        'medicare_ben': 'mcare_ben',
+        'medicaid_ben': 'mcaid_ben',
+        'SS': 'ss_ben',
+        'SNAP': 'snap_ben',
         'SLTX': 'e18400'
     }
     data = data.rename(columns=renames)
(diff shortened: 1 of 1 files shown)


martinholmer:
Thought we had all agreed to put the benefit data in its own file.  That’s what Amy was doing just a few days ago.  Has there been a change in plans?

andersonfrailey:
@martinholmer, my understanding is that we will include the first year of the benefit data in the CPS file, but all of the extrapolated data in a separate file.

@Amy-Xu am I interpreting what you're working on correctly or did I misunderstand?

Amy-Xu:
@andersonfrailey That seems to be the current arrangement in PR #108. @martinholmer We indeed want to put the extrapolated benefit data in a separate file, but it seems we never specified for the 2014 data. I lean toward putting it in the CPS file, but don't have a strong preference.

cc @hdoupe

martinholmer:
Sent from my iPhone

> On Nov 2, 2017, at 11:54 AM, Amy Xu <notifications@github.com> wrote:
> 
> @andersonfrailey That seems to be the current arrangement in PR #108. @martinholmer We indeed want to put the extrapolated benefit data in a separate file, but it seems we never specified for the 2014 data. I lean toward putting it in the CPS file, but don't have a strong preference.
> 
All the benefit data should be together, don’t you think Amy?

MattHJensen:
> All the benefit data should be together, don’t you think Amy?

Separating the base benefits data from cps.csv would significantly complicate the idea of a 'compatible_data' field in `current_law_policy.json` as some parameters would require multiple datasets. Why not include the benefits data with cps.csv but include the extrapolation-relevant auxiliary data in a separate file or files?

Amy-Xu:
@MattHJensen I would prefer what you suggested, but I guess what Martin @martinholmer  suggested was to create zero variables holders if no benefit reform is involved. Martin, is that right?

Then both are feasible?

andersonfrailey:
If there are no more issues with this PR, I would like to merge it at the end of the day.

There was some confusion about whether the benefit data will be included in the CPS file or separately, but I believe that has been addressed in the comments.

andersonfrailey:
There seem to be no more comments with this pull request so I am merging it.

PR #123, opened 2017-11-07
Add h_seq and ffpos Variables

andersonfrailey:
This PR completes a request from @MattHJensen to add the household sequence and the unique family identifier variables, `h_seq` and `ffpos`, respectively, to the CPS file. They are names in accordance with the [NBER](http://www.nber.org/cps/cpsmar2014.pdf) documentation. Using the two in combination, you can identify individual families in the CPS.

For example, you may have a household with `h_seq == 1` that contains two families. `ffpos` for the two families will be 1 and 2.

A corresponding Tax-Calculator PR is forth coming.

### DIFF
--- a/cps_data/finalprep.py
+++ b/cps_data/finalprep.py
@@ -48,12 +48,12 @@ def main():
         'BLIND_HEAD': 'blind_head',
         'BLIND_SPOUSE': 'blind_spouse',
         'HMIE': 'e19200',
-        # 'SSI': 'ssi_ben',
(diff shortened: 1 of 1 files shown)


MattHJensen:
Thanks @andersonfrailey! 

This request is from a journalist who uses tax-calculator and would like to match tax-calculator output with other demographic variables from the current population survey.

andersonfrailey:
Closed in favor of PR #124

Issue #119, opened 2017-10-24
Missing variables in CPS file


codykallen:
To allow people outside the OSPC to use B-Tax (beyond the limited capabilities of the webapp), they would need to be able to use the CPS instead. It is simple to modify B-Tax to call `Records.cps_constructor` instead of `Records()`, but the CPS is missing several critical variables for B-Tax. 

B-Tax requires marginal tax rates on 10 different variables from Tax-Calculator, but the CPS file is missing 4 of them:
`e02000`: Total Sch E income or loss
`e26270`: Income or loss from a partnership or S corporation
`p22250`: Short-term capital gain or loss
`p23250`: Long-term capital gain or loss

The missing variables `e02000` and `e26270` would cause incorrect calculations for METRs, METTRs and cost of capital. The missing variables `p22250` and `p23250` cause errors that prevent B-Tax from running (unless `test_run = True`, in which case the preset hardcoded values are used.

Would it be reasonable for someone to add impute these variables for the CPS file?

@MattHJensen @andersonfrailey @Amy-Xu @martinholmer

MattHJensen:
I believe this is an important issue for making the CPS file available on PolicyBrain, and it appears important for its own sake as well. '

My understanding was that we have imputations for these variables from John O'Hare, but it appears that may be incorrect (@andersonfrailey, could you chime in?)

If so, we may need to impute these variables onto the CPS file from the PUF. My view is that a simple imputations would be sufficient for now. 

adding to cc @jdebacker @hdoupe.

andersonfrailey:
We won't be able to impute the variables @codykallen mentioned using just the PUF data. We impute a few other variables in the PUF off of the CPS so we could expand that process to include the missing variables we need for B-Tax.

MattHJensen:
@andersonfrailey said:

> We won't be able to impute the variables @codykallen mentioned using just the CPS data. We impute a few other variables in the PUF off of the CPS so we could expand that process to include the missing variables we need for B-Tax.

We may need to do the opposite: impute _to_ the CPS _off of_ the PUF rather than off of the CPS to the PUF.

andersonfrailey:
@MattHJensen said:
> We may need to do the opposite: impute to the CPS off of the PUF rather than off of the CPS to the PUF.

You're right. My comment was poorly worded. We could use the same routine that we use for the of the deductions that are imputed onto the CPS: use the PUF to get beta coefficients that are used in the imputation on the CPS.

[...]

PR #127, opened 2017-11-16
Add e02000 and e26270 to CPS file

andersonfrailey:
This PR addresses issue #119 and adds `e02000` and `e26270` to the CPS file by splitting total self-employment income in the CPS using a method suggested by @codykallen.

The proportions used come from table 1.4 of the IRS [SOI tables](https://www.irs.gov/statistics/soi-tax-stats-individual-income-tax-returns-publication-1304-complete-report) for 2014. I used 2014 rather than the most recent file (2015) because we use the 2014 totals in all of our other projections and splitting functions. These numbers should be updated to reflect the 2015 numbers when we move to that file.

To find total self-employment income, I added net business or profession, net total rent and royalty, net partnership and S corporation, and net estate and trust. The proportions were then calculated the following way:

```
e00900 = net business or profession / total
e26270 = net partnership and S corporation / total
e02000_e26270 = (net total rent and royalty + net estate and trust) / total
```

@codykallen is that an accurate representation of what you suggested?

I'll merge upon review from @codykallen and @MattHJensen.

### DIFF
--- a/cps_data/finalprep.py
+++ b/cps_data/finalprep.py
@@ -20,8 +20,6 @@ def main():
         'JCPS21': 'e00200p',
         'JCPS31': 'e00200s',
         'ALIMONY': 'e00800',
-        'JCPS25': 'e00900p',
-        'JCPS35': 'e00900s',
         'JCPS28': 'e02100p',
         'JCPS38': 'e02100s',
         'UCOMP': 'e02300',
@@ -65,7 +63,6 @@ def main():
     # Use taxpayer and spouse records to get total tax unit earnings and AGI
     data['e00100'] = data['JCPS9'] + data['JCPS19']
     data['e00200'] = data['e00200p'] + data['e00200s']
-    data['e00900'] = data['e00900p'] + data['e00900s']
     data['e02100'] = data['e02100p'] + data['e02100s']
     # Determine amount of qualified dividends using IRS ratio
     data['e00650'] = data.e00600 * 0.7556
@@ -79,6 +76,8 @@ def main():
     # Split pentions and annuities using PUF ratio
     data['e01700'] = data['e01500'] * 0.1656
 
+    print 'Spliting Self-Employment Data'
+    data = split_se_income(data)
     print 'Applying deduction limits'
     data = deduction_limits(data)
     print 'Adding dependents'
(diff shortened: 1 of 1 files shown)


codykallen:
@andersonfrailey, this looks right to me. Good work on this.

MattHJensen:
@andersonfrailey how do the totals look vs. puf? Also, does anything need to happen in the extrapolation routine for these variables to be extrapolated properly?

andersonfrailey:
@MattHJensen, I'm about to do a comparison of the two and will post my results. There may be an effect on the stage 2 weighting process, but that is the one thing we can't currently change because we don't yet have the scripts to run the stage 2 process on the CPS.

andersonfrailey:
Initial pass over the data looks like this isn't correct. The weighted sum of each variable in the CPS is 30-40% of it's PUF counterpart. This holds for the years 2014-2027.

MattHJensen:
> Initial pass over the data looks like this isn't correct. The weighted sum of each variable in the CPS is 30-40% of it's PUF counterpart. This holds for the years 2014-2027.

Got it. What do you think the next step should be?

andersonfrailey:
@MattHJensen, in issue #119 you mentioned that a user said `e00900` in the CPS is roughly equivalent to the sum of IRS business and s/corp income. When I looked at it that doesn't seem to be the case.  Could you ask them how they came to their conclusion?

MattHJensen:
> @MattHJensen, in issue #119 you mentioned that a user said e00900 in the CPS is roughly equivalent to the sum of IRS business and s/corp income. When I looked at it that doesn't seem to be the case. Could you ask them how they came to their conclusion?

I'm asking. In the meantime, I'd suggest proceeding in the manner you think is appropriate.

martinholmer:
@andersonfrailey,  You need to change a variable name in this pull request per issue #129.

cps_data/finalprep.py
```diff
@@ -79,6 +76,8 @@ def main():
     # Split pentions and annuities using PUF ratio
     data['e01700'] = data['e01500'] * 0.1656
 
+    print 'Spliting Self-Employment Data'
```
MaxGhenis: Splitting

cps_data/finalprep.py
```diff
@@ -98,6 +97,24 @@ def main():
...
 
+def split_se_income(data):
+    """
+    Split self-employment data into e00900, e02000, and e26270 according to
+    IRS income data
+    """
+    total_se = data.JCPS25 + data.JCPS35
+    earnings_split_p = data.JCPS25 / total_se
+    earnings_split_s = 1. - earnings_split_p
+    data['e00900'] = total_se * 0.318327505
```
MaxGhenis: Sorry to pop in here, but I'm curious if the logic for producing constants like this 0.318 is documented outside the PR? It sounds like it's based on a few cells in the [2014 IRS data](https://www.irs.gov/pub/irs-soi/14in14ar.xls) ([spreadsheet](https://docs.google.com/spreadsheets/d/1LgWl68KveoGfjZbO8Aznpn6M_AsdqDBWs4LzCWVfW2g/edit?usp=sharing)), but I'm not totally clear on exactly which ones.

andersonfrailey: It's based on U9-W9 (business or profession), BA9-BC9 (total rent or royalty), BE9-BG9 (partnership and S corporation), and BI9-BK9 (estate and trust). But I'm looking at alternative methods to get at these variables so I wouldn't put too much time into working this out just yet.

PR #114, opened 2017-09-06
2011 PUF and 2016 CPS

andersonfrailey:
This PR makes the changes in TaxData to take us from a match of the 2009 PUF/2014 CPS to a match between the 2011 PUF/2016 CPS. Here's a quick summary of the changes:

1. Fixed a couple small bugs I found in the matching scripts and made some changes for PEP8 compliance.
2. Updated `cpsmar.py` to read in the raw 2016 CPS.
3. Updated `runmatch.py` to handle the changes between the 2009 and 2011 PUF files (the aggregate record that we drop was split into 4 separate records) and replace all missing variables with zero.
4. New weights, grow factors, and adjustment factors.
5. Updated `stage1-3.py` as necessary.

In issue #104 and PR #113 I discussed how the CyLP model we use in stage 2 no longer works with the most recent versions of scipy and numpy. I talked with @MattHJensen and in order to get this PR open as soon as possible I created a new Conda environment with earlier versions of Scipy and Numpy, 0.18.1 and 1.12.1, respectively, to run everything. I will be making that available as soon as I figure out how.

I put together a notebook to compare Tax-Calculator output between the current PUF and this new version to SOI reported data. That notebook can be seen [here](http://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/2011%20PUF%20Comparisons.ipynb). 

There are some noticeable differences between the current and new PUF, particularly in individual income tax and AMT liability, but those changes tend to bring the Tax-Calculator calculations closer to the SOI numbers (I'm comparing for the year 2014 because that is the most recent data from SOI). This is not the case for interest income and interest paid, both of which rose dramatically. I haven't yet figured out the cause of this.

If there is anything else you would like added to the notebook, let me know and I will do so.

cc @martinholmer @hdoupe @Amy-Xu

PR #132, opened 2017-12-04
Final updates to add benefit data to CPS

andersonfrailey:
This PR has the final modifications needed to add benefit data to the CPS.

@martinholmer and @MattHJensen, what are your thoughts on changing `n1821` to `n1820` in this PR and then releasing a new PUF with the change as well rather than waiting on my to finish analyzing the cause of the changes in PR #114?


andersonfrailey:
The changes in this PR have been addressed in #133 and #146. Closing.

PR #145, opened 2018-01-11
Add Fortran Code for CPS Stage 2

andersonfrailey:
John sent me the Fortran code he used to create the weights for the CPS file and this PR adds it to the cps_stage2 directory. There is an additional CSV file with state targets that needs to be added, but he has not sent it to me yet. I'll add that when he does.


andersonfrailey:
Latest commit adds the target file John provided me. I believe I should be able to run the Fortran code once I update the file paths in it. I'm looking for a Fortran compiler to confirm everything runs.

martinholmer:
@andersonfrailey,  I don't understand why the targets begin added in #145 are different from those in the `stage1/Stage_II_targets.csv` file.

I thought our methodology calls for the targets to be the same for the PUF and CPS files?  
But, for example, there is no "US Population" row in the file you are adding in PR #145.

I'm totally confused about how the two data sets are generated.  Can you provide some explanation?

andersonfrailey:
@martinholmer, the targets you mentioned are used in two ways: to create the PUF weights file, and to create the growth factors. We decided to use the same growth factors for both the CPS and PUF, but the weighting process uses different targets.

The PUF targets on the national level because we cannot accurately determine which state each record is from. The CPS, on the other hand, is targeted at the state level for weight creation. [State level IRS data](https://www.irs.gov/statistics/soi-tax-stats-historic-table-2) is used to establish targets for each individual state, and the linear programming model is run for each state/target pair.

The file I added contains the state level targets that John provided me. Does that clear things up a little?

martinholmer:
@andersonfrailey provided an explanation of PUF vs CPS weight derivation and then asked:
> The file I added contains the state level targets that John provided me. Does that clear things up a little?

Your explanation clears up much of my confusion.  Thanks.
But I still have three questions:

1. Is it correct to say: the CPS weights don't use the information in the `stage1/Stage_II_targets.csv` file, but the PUF weights are constructed using the information in that file?

2. Are the "US Population" targets in the `stage1/Stage_II_targets.csv` file used to construct the PUF weights in Stage II?

3. Are any population targets used to construct the CPS weights in Stage II?

MaxGhenis:
Do the state targets used for CPS add up to the national targets used for PUF?

andersonfrailey:
@martinholmer asked:
>Is it correct to say: the CPS weights don't use the information in the stage1/Stage_II_targets.csv file, but the PUF weights are constructed using the information in that file?

Yes.

>Are the "US Population" targets in the stage1/Stage_II_targets.csv file used to construct the PUF weights in Stage II?

Not directly. US population is not targeted in stage 2. The closest target to population targets we get are number of tax units (s006). The US population targets are used to create the `APOPN` factor, which is used throughout the stage 2 process.

>Are any population targets used to construct the CPS weights in Stage II?

It does not appear so. From what I've gathered reading the Fortran code included in this PR, population is not targeted, but number of returns are.

@MaxGhenis asked:
>Do the state targets used for CPS add up to the national targets used for PUF?

I haven't checked this yet, but they should, yes.

MaxGhenis:
[This spreadsheet](https://docs.google.com/spreadsheets/d/1iOwPCE8gIg7ay8V7zH939V_VsnINuknuimL14nB7Rsg/edit#gid=968696853) has CPS and PUF US-wide 2014 targets. It doesn't seem like many of these are comparable, and potentially comparable ones like returns by income are in different units, but maybe this could is useful down the line.

andersonfrailey:
Now that we've implemented a python version of the stage 2 weighting process, I'm going to close this PR.

Issue #149, opened 2018-02-12
What is relationship between XTOT and (nu18+n1820+n21)?


martinholmer:
Shouldn't the value of XTOT be equal to the value of (nu18+n1820+n21) for every filing unit in both the CPS data and in the PUF data?

martinholmer:
When I add a test of my expectations, I get this puzzling result:
```
______________________________ test_ubi_variables ______________________________

cps_path = '/Users/mrh/work/OSPC/tax-calculator/taxcalc/tests/../cps.csv.gz'

    @pytest.mark.one
    def test_ubi_variables(cps_path):
        """
        Ensure that the three UBI head-count variables add up to XTOT variable.
        """
        cpsdf = pd.read_csv(cps_path)
        xtot = cpsdf['XTOT']
        nsum = cpsdf['nu18'] + cpsdf['n1820'] + cpsdf['n21']
        if not np.allclose(xtot, nsum):
            print 'number of diffs', np.sum(xtot != nsum)
            print 'number xtot < nsum', np.sum(xtot < nsum)
            print 'number xtot > nsum', np.sum(xtot > nsum)
>           assert 'XTOT' == '(nu18+n1820+n21)'
E           AssertionError: assert 'XTOT' == '(nu18+n1820+n21)'
E             - XTOT
E             + (nu18+n1820+n21)

tests/test_cpscsv.py:140: AssertionError
----------------------------- Captured stdout call -----------------------------
number of diffs 17944
number xtot < nsum 679
number xtot > nsum 17265
============================= 441 tests deselected =============================
=================== 1 failed, 441 deselected in 6.81 seconds ==============
```

Are my expectations unreasonable?

Did I make a mistake in developing the new test?

I don't understand what's going on here.

@MattHJensen @Amy-Xu @andersonfrailey

andersonfrailey:
@martinholmer, my initial reaction is that test should work.

For households with multiple tax units, we check to see if any of the units should be combined. It's possible that there's an issue when they're combined that causes the discrepancy you've pointed out. I'll need to look into more before I can give you a definitive answer.

martinholmer:
@andersonfrailey,  Thanks for the quick feedback on issue #149.

martinholmer:
@andersonfrailey,  It would seem to me that resolving issue #149 should have a high priority because the accuracy of reforms that repeal benefits and add a UBI are open to question.

andersonfrailey:
[...]

PR #151, opened 2018-02-15
Fix XTOT and UBI variable error in CPS file

andersonfrailey:
This PR addresses a problem in the CPS file brought up by @martinholmer in issue #149. @martinholmer found that for some tax units, `XTOT` was not equal to the sum of `nu81`, `n1820`, and `n21`. After some digging, I found this was a problem in about 6% of all tax units and likely due to some individuals in the unit being miscounted.

To fix this, I added a check in the output macro to make sure `XTOT` equaled the sum of `nu81`, `n1820`, and `n21`. If it didn't, I reset all of the variables and looped through each member of the tax unit incrementing `XTOT` and the appropriate age variable as I went. 

I plugged in the new CPS file int Tax-Calculator to check for difference in the totals:

```
In [16]: nsums = c2.array('nu18') + c2.array('n1820') + c2.array('n21')

In [17]: (c2.array('XTOT') - nsums).sum()
Out[17]: 0
```

I also checked the combined income tax liability using the version of the CPS and there was no change from the current liability.

The SAS scripts in this PR show a lot more changes than I actually made. To edit/run them, I use a remote connection to a server. I'll then move them over to my Mac to make pull requests. I think this messes with line endings and spacing because most of the changes GitHub is finding are related to spacing and the end of lines.

### DIFF
--- a/cps_data/State Database - 2015/CPS Tax Units/2013/Programs/CPS-RETS13V5.SAS
+++ b/cps_data/State Database - 2015/CPS Tax Units/2013/Programs/CPS-RETS13V5.SAS
@@ -3,7 +3,7 @@
 			WHERE I MADE THE MODIFICATIONS, JUST SEARCH FOR "SHANE:". I ADDED 10 NEW
 			PERSON-LEVEL VARIABLES, BUT THIS TRANSLATED INTO 20 NEW TAX UNIT VARIABLES:
 			ONE FOR THE PRIMARY TAXPAYER AND ONE FOR THE SPOUSE. THE VARIABLES ARE FROM
-			YOUR EARLIER E-MAIL, PLUS I ADDED A COUPLE OF CENSUS IMPUTED TAX FIELDS. I 
+			YOUR EARLIER E-MAIL, PLUS I ADDED A COUPLE OF CENSUS IMPUTED TAX FIELDS. I
 			DON'T HAVE MUCH EXPERIENCE WITH THESE IMPUTED AMOUNTS.
 
 			1.	A_AGE	(ALREADY ON THE ORIGINAL TAX UNIT EXTRACT)
@@ -55,7 +55,7 @@
 ***                 * Repeat this process for each household member until all members   ***;
 ***                   of the household are represented in a tax unit record. A house-   ***;
 ***                   hold member can only be represented on one tax unit record.       ***;
-***                 * After all CPS tax units have been constructed, determine if the   ***; 
+***                 * After all CPS tax units have been constructed, determine if the   ***;
 ***                   unit is required to actually file a tax return as determined by   ***;
 ***                   filing threshholds in place for that tax year.                    ***;
 ***                 * Repeat this process for every household in the CPS until a SAS    ***;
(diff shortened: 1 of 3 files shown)


martinholmer:
@andersonfrailey,  When I use the new `cps.csv.gz` file in Tax-Calculator, there are **no differences** in test results (other than the new test that compares XTOT to the sum of the three UBI-related n* variables, which now passes).   I wasn't expecting no differences.  The new CPS data has different values for XTOT for some filing units, right?  If so, why do you suppose I didn't see any test results changes?

martinholmer:
@martinholmer said in PR #151:
> When I use the new `cps.csv.gz` file in Tax-Calculator, there are **no differences** in test results (other than the new test that compares XTOT to the sum of the three UBI-related n* variables, which now passes). I wasn't expecting no differences. The new CPS data has different values for XTOT for some filing units, right? If so, why do you suppose I didn't see any CPS test results changes?

After rereading @andersonfrailey's comments in issue #149, it now seems plausible that the CPS-related unit test results do not change when using the new `cps.csv.gz` file generated by PR #151.  Only a small fraction of filing units have a change in XTOT and those changes are mostly a change of one.  Given that unit test results are typically rounded to the nearest $0.1 billion, I guess it is not surprising that the Tax-Calculator unit test results did not change.

This data enhancement will be more important when users start to analyze reforms that repeal benefits and add a UBI and/or enhance refundable tax credits.

@andersonfrailey, thanks for the quick work on this.  I'm happy with the new `cps.csv.gz` file and will be incorporating it in a Tax-Calculator pull request.   I'm fine with you merging taxdata PR #151.

andersonfrailey:
Thanks for the review, @martinholmer. I'll get to work on making the same fix for the PUF file this afternoon.

martinholmer:
@andersonfrailey said in PR #151:
> To fix this, I added a check in the output macro to make sure XTOT equaled the sum of nu81, n1820, and n21. If it didn't, I reset all of the variables and looped through each member of the tax unit incrementing XTOT and the appropriate age variable as I went.

The way I read that statement is that some of the `XTOT` values are now different than before #151.
But that's not true when I tabulate the `XTOT` variables before and after these changes.

So, what changed?  Are the `nu18` or `n1820` or the `n21` values different for some filing units?

andersonfrailey:
> Are the nu18 or n1820 or the n21 values different for some filing units?

Yes. As I've dug into the issue more, I'm convinced that the values for `XTOT` were correct, and it was my code that created the n* variables causing the issues. When I was looking into the problem yesterday, `XTOT` was always reasonable, but the sum of n* would be somethin totally off like 7 or 10.

If you look at the (unweighted) sum of the `XTOT` array both before and after the fix, you get 1,029,271. And if you subtract the two arrays element wise the sum of the resulting array is zero. If you look at the sum of the n* variables, you get 1,012,676 before the fix and 1,029,271 after.

By adding in that check during the output macro, I effectively just made sure everyone already being counted in XTOT were also counted in the n* variables.

martinholmer:
@andersonfrailey,  thanks for the more detailed explanation about PR #151 in [this comment](https://github.com/open-source-economics/taxdata/pull/151#issuecomment-366083237).

Issue #131, opened 2017-11-27
why is n24 capped at 3 on cps.csv?


MattHJensen:
Why are there no values of n24 over 3 in the cps.csv file? SOI censors the PUF at 3, but why is the CPS-based file capped? 

cc @andersonfrailey @martinholmer @evtedeschi3

evtedeschi3:
Looks like it's [explicitly capped](https://github.com/open-source-economics/taxdata/blob/master/cps_data/finalprep.py) in the `finalprep.py` file to match what would be seen in the `puf.csv`. See lines 144-153:

```
    # Count number of children eligible for child tax credit
    # Max of three to mach PUF version of n24
    age1 = np.where((data.ICPS03 > 0) & (data.ICPS03 <= 17), 1, 0)
    age2 = np.where((data.ICPS04 > 0) & (data.ICPS04 <= 17), 1, 0)
    age3 = np.where((data.ICPS05 > 0) & (data.ICPS05 <= 17), 1, 0)
    age4 = np.where((data.ICPS06 > 0) & (data.ICPS06 <= 17), 1, 0)
    age5 = np.where((data.ICPS07) > 0 & (data.ICPS07 <= 17), 1, 0)
    n24 = age1 + age2 + age3 + age4 + age5
    n24 = np.where(n24 > 3, 3, n24)
    data['n24'] = n24
```

This is interesting because here is an area where the CPS has more reliable data than the PUF. Why is the CPS being censored here? Since the PUF is matched with the CPS, couldn't/shouldn't the CPS be used to fill in detail for PUF tax units likely to have more than 3 adults?

andersonfrailey:
It was capped to match the PUF file. I can update the final prep scripts to change this when I get back in the office.

MaxGhenis:
Just checking in on this, as I'm working on a CTC analysis.

andersonfrailey:
Thanks for bringing this back up @MaxGhenis. I'm working on an update to the CPS file that should be ready in the next day or two. I'll address this in that update as well.

andersonfrailey:
This issue has been resolved in PR #161. Closing.

PR #161, opened 2018-03-02
Add Additional Imputed Benefits

andersonfrailey:
This PR adds new variables imputed by C-TAM: WIC, Housing benefits, TANF, and Unemployment insurance. The first three have not previously been in the CPS file, and the imputed unemployment insurance will replace `e02300`. I've removed housing and unemployment assistance from the `benefitprogram.csv` file. It also removes the cap on `n24` in the CPS file, to address issue #131.

The reason I have labeled this "in progress" is there are two items in `benefitprograms.csv` that need to be evaluated. `Family support payments to States and TANF` and `Supplemental feeding programs (WIC and CSFP)` contain programs that have now been imputed, but also other programs that have not. @Amy-Xu and I have not yet decided what the best way to handle this is yet. One thought is to simply subtract the weighted sum of the imputed TANF and WIC variables from the amount in `benefitprograms.csv`, but I'd appreciate other opinions if there are any.


andersonfrailey:
Latest commit removed TANF and WIC from the total `other_ben` before it is distributed to avoid double counting.

cps_data/finalprep.py
```diff
@@ -381,8 +384,11 @@ def benefits(data, other_ben):
     data['ssi_ben'] = data['ssi_ben'].astype(np.int32)
     data['snap_ben'] = data['snap_ben'].astype(np.int32)
     data['vet_ben'] = data['vet_ben'].astype(np.int32)
+    data['tanf_ben'] = data['tanf_ben'].astype(np.int32)
+    data['wic_ben'] = data['wic_ben'].astype(np.int32)
+    data['housing_ben'] = data['housing_ben'].astype(np.int32)
     data['e02400'] = data['e02400'].astype(np.int32)
-    data['e02300'] = data['e02300'].astype(np.int32)
+    data['e02300'] = data['ui_ben'].astype(np.int32)
```
Amy-Xu: @andersonfrailey So are we replacing e02300 with the imputed unemployment compensation? In other words, we're using imputed imputed UI for tax calculations as well? I thought it was the other way around but total fine with this if the tax liabilities look ok. @MattHJensen Do you think it's fine  to use imputed benefit to replace e02300? Anderson previously checked the total of imputed UI is 93% of e02300.

cps_stage3/extrapolation.py
```diff
@@ -203,7 +203,7 @@ def _read_data(self, growth_rates, cps_benefit, cps_weights,
         for benefit in benefit_names:
 
             # create benefit targets
-            benefit_2014 = (cps_benefit[benefit] * cps_benefit.s006).sum()
+            benefit_2014 = (cps_benefit[benefit] * cps_benefit.WT).sum()
```
Amy-Xu: I'm not super familiar with the previous version of this script, so just want to check: is this a switch from s006 to WT for every year or just a renaming thing? In other words, are these two variables completely equivalent for every year?

andersonfrailey: Yes. When we create the `s006` variable for the final CPS value we're just renaming the `WT` column. I'm actually still trying to figure out why it could run at all using `s006` instead of `WT`.

andersonfrailey:
Last commit adds the edits to the SAS scripts I made

andersonfrailey:
If there are no more comments, I'll merge this at the end of the day.

martinholmer:
@andersonfrailey, In PR #161 you added four new benefits (ui, housing, wic, tanf), but removed only two benefits (ui, housing) from the list of other benefits in the `cps_data/benefitprograms.csv` file.
Why didn't you remove wic and tanf from the list of other benefits?  Seems like the current state of affairs involves some double counting of benefits.

@Amy-Xu

andersonfrailey:
@martinholmer I left in the line items for `Supplemental feeding programs (WIC and CSFP)` and `Family support payments to States and TANF` because, while they each contain TANF and WIC, those are not the only programs included in each line item. To account for double counting I subtracted the weighted sum of imputed TANF and WIC benefits from the sum of `other_ben` before distributing the non-modeled benefits.

martinholmer:
@andersonfrailey said about PR #161:
>  I left in the line items for Supplemental feeding programs (WIC and CSFP) and Family support payments to States and TANF because, while they each contain TANF and WIC, those are not the only programs included in each line item. To account for double counting I subtracted the weighted sum of imputed TANF and WIC benefits from the sum of other_ben before distributing the non-modeled benefits.

OK.  So, the new CPS data files are all three?  
`cps.csv.gz`
`cps_weights.csv.gz`
`cps_benefits.csv.gz`
From which directories in the taxdata repository?

andersonfrailey:
@martinholmer, only `cps.csv.gz` in the `cps_data` directory and `cps_benefits.csv.gz` in the `cps_stage3` directory need to be updated. `cps_weights.csv` has not changed.

martinholmer:
@andersonfrailey said:
> only cps.csv.gz in the cps_data directory and cps_benefits.csv.gz in the cps_stage3 directory need to be updated. cps_weights.csv has not changed.

Got it!  Thanks.

Issue #164, opened 2018-03-03
Various age fields include end points when names and documentation indicate otherwise


MaxGhenis:
Documentation and variable name structure for various age fields indicate they should not include the end point. I see four contradictions in the code on this front:

* **`nu05`** includes 5-year-olds based on the [finalprep.py code](https://github.com/open-source-economics/taxdata/blob/801164146006ab9c9b7f8872474b9a6cdec9ada4/cps_data/finalprep.py#L135).
* **`nu13`** includes 13-year-olds based on the [finalprep.py code](https://github.com/open-source-economics/taxdata/blob/801164146006ab9c9b7f8872474b9a6cdec9ada4/cps_data/finalprep.py#L127)
* **`elderly_dependent`** is described as "over 65 years old" in the documentation, but includes 65-year-olds based on the [finalprep.py code](https://github.com/open-source-economics/taxdata/blob/801164146006ab9c9b7f8872474b9a6cdec9ada4/cps_data/finalprep.py#L155). As an aside, it would be helpful if the [user documentation](http://open-source-economics.github.io/Tax-Calculator/) specified the age threshold for this variable.
* **`n24`** has a related issue: only children [*under* age 17](https://www.irs.gov/pub/irs-pdf/p972.pdf) are eligible for the CTC, but [finalprep.py](https://github.com/open-source-economics/taxdata/blob/801164146006ab9c9b7f8872474b9a6cdec9ada4/cps_data/finalprep.py#L145) includes 17-year-olds in `n24`.

`n24` aside, I'm not sure if these are used elsewhere, and how that affects whether the variable names or the logic should be changed.

Relevant links:
* https://github.com/open-source-economics/Tax-Calculator/issues/1669.
* http://open-source-economics.github.io/Tax-Calculator/
* https://github.com/open-source-economics/taxdata/blob/master/cps_data/cps_file_doc.md

MaxGhenis:
A process tweak that could help avoid this is to use exclusively either `<`/`>` or `<=`/`>=` in code. finalprep.py includes endpoints for these variables, but just below them it doesn't with the `f2441` and `EIC` variables.

martinholmer:
@andersonfrailey,
Where do we stand on correcting these bugs in the age variables in the `cps.csv` file?
Are the same kind of errors also present in the `puf.csv` file?

@MattHJensen @MaxGhenis

andersonfrailey:
[...]

PR #172, opened 2018-03-19
Fix Age Variables

andersonfrailey:
This PR addresses concerns raised by @MaxGhenis in issue #164. I've corrected the `nu05`, `nu13`, and `n24` variables in the CPS file, as well as modified the scripts to create the PUF. Interestingly, I got the same MD5 output from the PUF after rerunning the scripts with the included modification.

cc @martinholmer

### DIFF
--- a/cps_data/cps_file_doc.md
+++ b/cps_data/cps_file_doc.md
@@ -148,9 +148,9 @@ The student loan interest deduction is capped at $2,500, and IRA contributions
 are limited at either $6,500 or $5,500 depending on the age of the head of the
 tax unit.
 
-We then count the number of dependents under 5 and 13 or over 65 years old.
-Finally, the number of children qualified for the child tax credit, childcare
-credit, and earned income credit are counted for each tax unit.
+We then count the number of dependents under 5 years old, under 13, and 65 or
+older. Finally, the number of children qualified for the child tax credit,
+childcare credit, and earned income credit are counted for each tax unit.
 
 The last step in the final preparations is adjusting the distribution of interest
 income, ordinary and qualified dividends, and self-employment income using a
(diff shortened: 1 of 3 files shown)


MaxGhenis:
> Interestingly, I got the same MD5 output from the PUF after rerunning the scripts with the included modification.

Are the source fields like `ICPS03` ints?

andersonfrailey:
@MaxGhenis, no. We've written the Python scripts so that those variables are incremented while dependents are added to the tax unit, rather than after the fact.

martinholmer:
@andersonfrailey, using the new `puf.csv` file generated by the changes in #172, will the following Tax-Calculator test pass (after removing the xfail decorator)?  If not, why does the test fail?  Is the test logic incorrect?  Or are there still problems with the age count variables in the `puf.csv` file?
```
@pytest.mark.requires_pufcsv
@pytest.mark.xfail
def test_ubi_n_variables(puf_path):
    """
    Ensure that the three UBI n* variables add up to XTOT variable,
    recognizing that XTOT values are often capped in the IRS-SOI PUF,
    so that XTOT < NSUM might not indicate any data inconsistency.
    """
    pufdf = pd.read_csv(puf_path)
    xtot = pufdf['XTOT']
    nsum = pufdf['nu18'] + pufdf['n1820'] + pufdf['n21']
    if not np.sum(xtot > nsum) == 0:
        print('number xtot > nsum is:', np.sum(xtot > nsum))
        assert 'XTOT' <= '(nu18+n1820+n21)'
```

andersonfrailey:
@martinholmer, that will still fail. I haven't been able to address that issue just yet.

andersonfrailey:
I've updated this PR to include the last few PR's that have been merged. As I mentioned yesterday, I will be offline until the 10th, but this PR is ready for review and can be merged by one of the other maintainers if there are no issues with it while I'm gone.

Issue #176, opened 2018-03-30
The cps_weights.csv and puf_weights.csv files have different formats


martinholmer:
The non-standard format of `cps_weights.csv` is causing substantial problems in Tax-Calculator.
What's the timeline for making the `cps_weights.csv` file have the same format as the `puf_weights.csv` file?

Here is the format difference:
```
$ gunzip -k puf_weights.csv.gz 
$ head -2 puf_weights.csv
WT2011,WT2012,WT2013,WT2014,WT2015,WT2016,WT2017,WT2018,WT2019,WT2020,WT2021,WT2022,WT2023,WT2024,WT2025,WT2026,WT2027
147096,146655,149097,148523,150406,152774,155407,157834,159967,161850,163762,165601,167454,169323,171220,173147,175074

$ gunzip -k cps_weights.csv.gz 
$ head -2 cps_weights.csv
SEQUENCE,WT2015,WT2016,WT2017,WT2018,WT2019,WT2020,WT2021,WT2022,WT2023,WT2024,WT2025,WT2026,WT2027
100,23002,23312,23627,23946,24269,24597,24929,25265,25606,25952,26302,26658,27017
```
The `cps_weights.csv` file is missing `WT2014` and has instead a mysterious `SEQUENCE` variable.

For an earlier discussion of this issue, see issue #125.

andersonfrailey:
@martinholmer, I can get a PR open to drop the sequence variable now. As for the 2014 weights column, I can copy over the weights from the CPS file, but that seems somewhat redundant. I still haven't received a definitive answer from John about the duplicate weights issue I brought up in issue #125.

martinholmer:
@andersonfrailey said:
> I can get a PR open to drop the sequence variable now. As for the 2014 weights column, I can copy over the weights from the CPS file, but that seems somewhat redundant. I still haven't received a definitive answer from John about the duplicate weights issue I brought up in issue #125.

I don't believe the `s006` values in the `cps.csv.gz` file.  So, don't do anything now.

Instead, focus on figuring out what the 2014 weights **should** be and then update the weights file.

andersonfrailey:
@martinholmer got it. I'll go ahead and remove the sequence variable and keep working on the 2014 weights issue.

martinholmer:
@andersonfrailey said in issue #176 conversation:
[...]

PR #184, opened 2018-04-24
Add 2014 column to weights

andersonfrailey:
This PR adds the 2014 weights from the CPS file to the CPS weights file to address and issue raised by @martinholmer in #176.

### DIFF
--- a/cps_stage2/finalprep.py
+++ b/cps_stage2/finalprep.py
@@ -6,5 +6,10 @@ weights = pd.read_csv('cps_weights_raw.csv.gz', compression='gzip')
 weights = weights.drop('SEQUENCE', axis=1)
 weights *= 100.
 weights = weights.round(0).astype('int64')
-weights.to_csv('cps_weights.csv', index=False)
+# add weights from CPS as 2014 value
+cps = pd.read_csv('../cps_data/cps.csv.gz')
+cps.s006 *= 100
+df = pd.concat([cps.s006, weights], axis=1)
+df = df.rename({'s006': 'WT2014'}, axis=1)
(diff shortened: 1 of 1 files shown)


martinholmer:
@andersonfrailey, The changes to `finalprep.py` in PR #184 look good except for one thing.  Here's the code:
```
weights = weights.drop('SEQUENCE', axis=1)
 weights *= 100.
 weights = weights.round(0).astype('int64')
# add weights from CPS as 2014 value
cps = pd.read_csv('../cps_data/cps.csv.gz')
cps.s006 *= 100
df = pd.concat([cps.s006, weights], axis=1)
df = df.rename({'s006': 'WT2014'}, axis=1)
df.to_csv('cps_weights.csv', index=False)
```
Looks to me as if the `WT2014` values are not going to be integers.
Shouldn't there be a statement like the following one just before the `pd.concat` call?
```
cps = cps.round(0).astype('int64')
```

andersonfrailey:
@martinholmer asked:

>Looks to me as if the WT2014 values are not going to be integers.
Shouldn't there be a statement like the following one just before the pd.concat call?

An additional conversion statement is not needed here because I'm pulling the weights from the final CPS file where they have already been converted to integers, rather than from the raw CPS file.

martinholmer:
@andersonfrailey said:
> An additional conversion statement is not needed here because I'm pulling the weights from the final CPS file where they have already been converted to integers, rather than from the raw CPS file.

You're right, I forgot about that.  
But given what you say, haven't the 's006' weights already been multiplied by 100?

andersonfrailey:
@martinholmer, they have not been which is an error in the final prep scripts. I'll update this PR accordingly.

PR #189, opened 2018-05-02
Fix Charitable Deduction Values

andersonfrailey:
I noticed that some tax units in the CPS had negative charitable deduction values. This PR changes the logic for how we calculate charitable deduction values so that they adhere to the "half of AGI" cap, but will always be at least zero.


andersonfrailey:
If there are no comments, I'll merge at the end of the day.

PR #177, opened 2018-03-30
remove sequence column from weights

andersonfrailey:
This PR partially resolves issue #176 by removing the `SEQUENCE` column from the final CPS weights file.

PR #190, opened 2018-05-03
Change Path for CPS Benefit Extrapolation Weights

andersonfrailey:
The CPS benefit extrapolation scripts currently read in the cleaned weights and use the `SEQUENCE` column that was dropped in PR #177. This PR updates them to use the raw weights file instead. There is no significant change in extrapolated benefit value.

cc @hdoupe


andersonfrailey:
If there are no comments, I'll merge this at the end of the day.

PR #191, opened 2018-05-04
Use minimal environment.yml file

hdoupe:
This PR proposes a simple `environment.yml` file that can easily be used and built on in the future. The file proposed here probably does not have all of the packages required to run everything in taxdata. However, I think we can get there by using this file, adding packages, and updating package versions as needed. 

I'm having trouble installing cylp, but the rest of the packages install OK. It may be worth adding separate documentation for installing cylp since it requires several extra steps. @andersonfrailey I understand that you are trying to find an alternative to cylp. So, I trust your judgement in adding this documentation or waiting until a more permanent linear programming solution is found.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ dependencies:
 - pandas >=0.22.0
 - bokeh >=0.12.3
 - pytest
-pip:
-- ipython
-- ipykernel
-- cylp==0.7.1
+- pip:
+  - ipython
+  - ipykernel
+  - cylp==0.7.1
--- a/environment.yml
+++ b/environment.yml
@@ -1,6 +1,6 @@
 name: taxdata-dev
 dependencies:
-- python=2.7.15
+- python>=2.7.14
 - numpy >=1.12.1
 - pandas >=0.22.0
 - bokeh >=0.12.3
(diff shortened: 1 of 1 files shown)


hdoupe:
@andersonfrailey numpy and pandas are pinned to the previous versions. I'm still unable to install cylp.

martinholmer:
@hdoupe, what's the reason for exact pinning of numpy and pandas.
We don't need to do this in Tax-Calculator.  Why do we need to do it in taxdata?

andersonfrailey:
@martinholmer, we need to pin pandas, numpy, and scipy (Hank will be adding Scipy to the PR shortly) because the linear programming model used in the PUF weights creation doesn't work with later versions of those packages. Another reason why we're searching for a different LP package.

martinholmer:
@andersonfrailey said:
> we need to pin pandas, numpy, and scipy (Hank will be adding Scipy to the PR shortly) because the linear programming model used in the PUF weights creation doesn't work with later versions of those packages. Another reason why we're searching for a different LP package.

Thanks for the explanation.

andersonfrailey:
Thanks, @hdoupe! Merging.

PR #193, opened 2018-05-04
Add Tests

andersonfrailey:
This PR starts the process of adding a test suite to this repo. I've added tests to check the max and minimum values for each variable, relative values between certain variables, and moved tests written by @hdoupe a while back for the benefit extrapolation.

Not everything is passing right now. One problem is negative values for charitable giving in the CPS, which is addressed in PR #189. Another is in the benefit extrapolation tests. @hdoupe is looking into that.

cc @martinholmer


martinholmer:
@andersonfrailey,  I got only as far as the second item in the `records_metadata.json` file:
```
 "EIC": {
    "type": "int",
    "desc": "number of EIC qualifying children (range: 0 to 3)",
    "form": {
      "2013-2016": "1040 Sch EIC"
    },
    "availability": "taxdata_puf, taxdata_cps",
    "range": {
      "min": 0,
      "max": 9e+99
    }
  },
```
There seems to be a logical inconsistency with respect to the valid range of values: not only is 9e99 not an integer value, but 9e99 is in conflict with the `desc`.

andersonfrailey:
@martinholmer, you're right. I used a script to create that JSON file to save myself some time and have been going through it to make sure the values make sense. I must've missed that one.

andersonfrailey:
I updated some of the min/max values. What are peoples thought's on max values for variables such as `nu18` and `n21`? The highest value in any file right now for one of the age variables is 12, but there's no real way of confirming that will be the case in future releases.

martinholmer:
@andersonfrailey said:
> I updated some of the min/max values. What are peoples thought's on max values for variables such as `nu18` and `n21`? The highest value in any file right now for one of the age variables is 12, but there's no real way of confirming that will be the case in future releases.

Why is maximum age set at 85?

On the maximum count by age group, how about leaving them at the maximum of that variables (in the current data set).  This would be the strongest test that would reveal the most unintended mistakes in data preparation.  When new data files produce a case or two that have an even higher count, the test maximum value can be raised after double-checking the validity of the cases with the higher counts.

andersonfrailey:
@martinholmer 
>Why is maximum age set at 85?

85 is the age cap in the CPS and we pull all of our age data from the CPS.

>On the maximum count by age group, how about leaving them at the maximum of that variables (in the current data set). This would be the strongest test that would reveal the most unintended mistakes in data preparation. When new data files produce a case or two that have an even higher count, the test maximum value can be raised after double-checking the validity of the cases with the higher counts.

This makes sense to me. I'll make those changes.

andersonfrailey:
>85 is the age cap in the CPS and we pull all of our age data from the CPS.

I forgot that in the PUF we do a little work with the age of a spouse in a tax unit so I've upped the maximum value for `age_spouse`

I also changed how we define the `elderly_dependent` variable in both the CPS and PUF scripts so that it is a 1/0 indicator, consistent with the documentation, rather than a counter.

Right now, there is one other issue I'm running into. In about 6% of PUF records, `n24` exceeds `nu18`. In these cases, the tax units that get matched together are the same size, but the ages of the individuals in the unit differ. This could probably be fixed by creating even smaller groups to match within, but that could make our matching accuracy worse in other areas by pairing up tax units that aren't particularly alike except in the ages of the people in the unit. Would be interested in hearing what others have to think about this potential trade off.

martinholmer:
@andersonfrailey asked in taxdata PR #193:
> Right now, there is one other issue I'm running into. In about 6% of PUF records, n24 exceeds nu18. In these cases, the tax units that get matched together are the same size, but the ages of the individuals in the unit differ. This could probably be fixed by creating even smaller groups to match within, but that could make our matching accuracy worse in other areas by pairing up tax units that aren't particularly alike except in the ages of the people in the unit. Would be interested in hearing what others have to think about this potential trade off.

This is just one of many asserts in the `relationships(data, dataname)` function.  What I would suggest is to weaken the test **only** for PUF dataname and **only** for the `assert np.all(data['n24'] <= data['nu18']), m` statement.  One way to do this would be something like this (untested) code:
```
if dataname == 'CPS':
    assert np.all(data['n24'] <= data['nu18']), m
else:  # PUF data
    if not np.all(data['n24'] <= data['nu18']):
        # implement weaker test here
```
The weaker PUF test should definitely include as a comment your above analysis of the situation in order to explain why the test is being weakened.  

The nature of the weaker testing depends on the nature of the assert failures.  

I would test for the number of assert failures (the unweighted count corresponding to your 6.4% figure, which I'm assuming is weighted).  You want the weaker test to warn you if these two PUF variables get out of line for more filing units in future PUF versions.

And I would also include the following.  For the failures, is `n24` always just one greater than `nu18`?  If so, you could weaken the assert using that information.  For **only** those that fail the original assert, they should pass this assert test: `np.all(data['n24'] <= data['nu18'] + 1)`.  If sometimes greater than one, the add more than one to `nu18`, but only for those who fail the original assert.

Does this make sense?

andersonfrailey:
Thanks for your suggestions @martinholmer. They make sense to me and I'll work on implementation today.

andersonfrailey:
Didn't realize I never pushed my changes. @martinholmer is what I've done about what you had in mind?

andersonfrailey:
@martinholmer, I've finished the initial tests for the files themselves. I don't think @hdoupe has had time to work on the benefit extrapolation tests though. I could update this PR to temporarily remove those tests so we can move forward.

I think it would be a good idea to set up travis for this repo to run the tests. I don't know how that works right now, but I can figure it out.

In the meantime, I'll edit the README file to instruct contributors to run the tests before opening a PR.

martinholmer:
@andersonfrailey,  I can't pass the tests when on PR#193.

When I use `./gitpr 193` to get #193  on my computer, running the `pytest` command produces three errors:
```
iMac2:taxdata mrh$ git branch
  master
* pr-193
iMac2:taxdata mrh$ pytest
============================= test session starts ==============================
platform darwin -- Python 2.7.14, pytest-3.4.2, py-1.5.2, pluggy-0.6.0
rootdir: /Users/mrh/work/OSPC/taxdata, inifile:
plugins: xdist-1.22.1, forked-0.2
collected 3 items                                                              

tests/test_data.py F                                                     [ 33%]
tests/test_extrapolation.py FF                                           [100%]
```
I guess the extrapolation test failures are to be expected given @hdoupe hasn't finished that benefits work.
Why don't you mark them as expected to fail using the `@pytest.mark.xfail` decorator?

Here are the details of the data test failure I get.
```
=================================== FAILURES ===================================
__________________________________ test_data ___________________________________

cps =         age_head  age_spouse  e00200p  e00900p  e02100p  e00200s  e00900s  \
0...   0       0       11          0  134271        0  

[456465 rows x 67 columns]
puf =         DSI   e00200   e00300   e00400   e00600   e00650   e00700  e00800  \
0....0      0.0       0.00      0.00        1       0  

[215525 rows x 87 columns]
metadata = {'DSI': {'availability': 'taxdata_puf, taxdata_cps', 'desc': '1 if claimed as dependent on another return; otherwise 0...te, 4=household-head, 5=widow(er)]', 'form': {'2013-2016': '1040 lines 1-5'}, 'range': {'max': 5, 'min': 1}, ...}, ...}

    def test_data(cps, puf, metadata):
        min_max(cps, metadata, 'cps')
>       min_max(puf, metadata, 'puf')

tests/test_data.py:67: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

data =         DSI   e00200   e00300   e00400   e00600   e00650   e00700  e00800  \
0....0      0.0       0.00      0.00        1       0  

[215525 rows x 87 columns]
meta = {'DSI': {'availability': 'taxdata_puf, taxdata_cps', 'desc': '1 if claimed as dependent on another return; otherwise 0...te, 4=household-head, 5=widow(er)]', 'form': {'2013-2016': '1040 lines 1-5'}, 'range': {'max': 5, 'min': 1}, ...}, ...}
dataname = 'puf'

    def min_max(data, meta, dataname):
        """
        Test for ensuring all variables are within their logical minimum and
        maximum
        """
        for var in meta.keys():
            availability = meta[var]['availability']
            min_value = meta[var]['range']['min']
            max_value = meta[var]['range']['max']
            in_data = True
            if dataname not in availability:
                in_data = False
            if in_data:
                m = '{}-{} contains values less than min value'.format(dataname,
                                                                       var)
                assert np.all(data[var] >= min_value), m
                m = '{}-{} contains values greater than max value'.format(dataname,
                                                                          var)
>               assert np.all(data[var] <= max_value), m
E               AssertionError: puf-elderly_dependent contains values greater than max value
E               assert False
E                +  where False = <function all at 0x103bf12a8>(0         0\n1         0\n2         0\n3         0\n4         0\n5         0\n6     ...0\n215523    0\n215524    0\nName: elderly_dependent, Length: 215525, dtype: int64 <= 1)
E                +    where <function all at 0x103bf12a8> = np.all

tests/test_data.py:22: AssertionError
```
What's the problem?

andersonfrailey:
@martinholmer, the test failure you're getting is related to the change in `elderly_dependent` logic. I'll update that in a new PR as you suggested and everything should pass afterwards.

andersonfrailey:
@martinholmer, I've updated the description to reflect the change in max/min values.

martinholmer:
@andersonfrailey,  Can you fix the several `max` value problems in the `tests/records_metadata.json` file in PR #193 and mark the unfinished benefit extrapolation tests with a `@pytest.mark.skip` decorator, and then merge #193?

When you do that, will this finish all the outstanding work we scheduled recently for improving the CPS and PUF data used by Tax-Calculator?

martinholmer:
@andersonfrailey,  These changes look great.  Why don't you go ahead and merge #193.

PR #194, opened 2018-05-18
fix elderly dependents variable

andersonfrailey:
This PR changes the logic used to create the `elderly_dependent` variable so that it will be either a 0 (no elderly dependents) or 1 (elderly dependent(s)), as the record metadata indicates.


martinholmer:
@andersonfrailey,  can you send me via private email the newest version of `puf.csv`?
The data tests are still failing.

PR #197, opened 2018-05-22
Cap EIC at 3

andersonfrailey:
This PR caps the EIC value in the CPS at 3 so that it works properly with Tax-Calculator.

cc @martinholmer


andersonfrailey:
If there are no comments I'll merge this at the end of the day.

PR #188, opened 2018-05-02
Fix PUF XTOT and N* Values

andersonfrailey:
This PR addresses issue #149 by @martinholmer. I've modified our statistical matching scripts to ensure that CPS tax units and PUF tax units will be matched in a way that ensures `XTOT` will be less than or equal to the sum of `nu18`, `n1820`, and `n21`. I've also created new weights and adjustment ratios.

I ran Tax-Calculator with the new PUF and saw only trivial changes in combined tax liability:

| Year | Current               | New                   | Diff                | % Change | 
|------|-----------------------|-----------------------|---------------------|----------| 
| 2018 | $2,669,810,142,992.14 | $2,663,304,172,483.77 | $-6,505,970,508.37  | -0.24    | 
| 2019 | $2,777,417,911,535.05 | $2,770,249,336,121.81 | $-7,168,575,413.24  | -0.26    | 
| 2020 | $2,883,297,994,543.28 | $2,875,747,663,018.96 | $-7,550,331,524.32  | -0.26    | 
| 2021 | $2,999,523,877,167.46 | $2,991,317,614,985.42 | $-8,206,262,182.04  | -0.27    | 
| 2022 | $3,128,265,046,067.50 | $3,120,106,696,584.30 | $-8,158,349,483.21  | -0.26    | 
| 2023 | $3,269,568,279,451.84 | $3,260,975,381,222.34 | $-8,592,898,229.50  | -0.26    | 
| 2024 | $3,417,521,728,848.76 | $3,408,179,840,379.17 | $-9,341,888,469.60  | -0.27    | 
| 2025 | $3,570,882,409,629.56 | $3,561,454,490,070.37 | $-9,427,919,559.19  | -0.26    | 
| 2026 | $3,966,239,038,871.07 | $3,953,748,425,282.18 | $-12,490,613,588.89 | -0.31    | 
| 2027 | $4,145,087,918,985.00 | $4,131,299,747,498.82 | $-13,788,171,486.18 | -0.33    | 

cc @Amy-Xu

PR #199, opened 2018-05-24
Fix Bug in PUF Weights

andersonfrailey:
PR #188 introduced a bug in the PUF weights that resulted in tax liabilities, as calculated by taxcalc, to be ten times larger than they should have been. I've recreated the weights and ratios files to fix this error.

cc @martinholmer


andersonfrailey:
@martinholmer, the code wasn't actually the issue here. I ran the exact same stage 2 scripts, but this time the weights were reasonable. My best intuition here is that either that specific copy of `puf_weights.csv.gz` was corrupted somehow or the `cps-matched-puf.csv` file it read in was wonky and resulted in bad weights.

I re-ran the PUF creation scripts yesterday which overwrote the previous version of `cps-matched-puf.csv` so I can't confirm the later theory, but that is the only thing that changed between the creation of this version of the weights and the one merged in PR #188.

martinholmer:
@andersonfrailey,  Thanks for the explanation and fix of the PUF weights and ratios.

PR #200, opened 2018-05-25
CPS Weights Creation Script

andersonfrailey:
This PR makes all of the changes necessary to create the CPS weights file from scratch.

**First**, I’ve added all of the python scripts I wrote to the `cps_stage2` folder as well as a notebook that does the same work. Because we’ll be able to create them from scratch now, I deleted `cps_weights_raw.csv.gz` and `finalprep.py`.

I used the [PuLP](https://pythonhosted.org/PuLP/index.html) optimization package for the linear programming model. Like the CyLP package we use to create the PUF weights, it doesn’t seem like it is being updated all that often, but it’s stable, works with Python 3.6, and will allow us to un-pin the versions of numpy, pandas, and scipy in the tax data environment if/when we update the PUF weights creation process from using CyLP to using PuLP.

The scripts take about 3.5 hours to run. This is primarily due to how long it takes PuLP to set up the LP model. There may be a way to speed things up, but I figure that can wait until a later PR because we (hopefully) won’t be recreating the weights file all that often.

**Second**, I created a new `cps_stage1` directory. This contains SOI estimates used to create the targets, stage 1 type factors, the final stage 2 targets, and a python script to produce the later two files.

I added new SOI estimates because the targets here are slightly different from those used to create the PUF weights. First, interest income now includes both taxable and tax-exempt interest. This is because the CPS interest totals include both and we split into taxable/tax-exempt during final prep. Second, we use different wage bins. This is because the CPS doesn’t have the best data on those with extremely high or extremely low AGI. The bins used here are:
1. AGI <= $10,000
2. $10,000 < AGI <= $20,000
3. $20,000 < AGI <= $30,000
4. $30,000 < AGI <= $40,000
5. $40,000 < AGI <= $50,000
6. $50,000 < AGI <= $75,000
7. $75,000 < AGI <= $100,000
8. AGI > $100,000

The new stage 1 factors are created using the stage 2 targets from the PUF extrapolation, but with the factors based at 2014 rather than 2011. 

**Third**, I renamed the directory `stage1` to `puf_stage1`.

You’ll notice that I’ve targeted at the aggregate level rather than state level. I originally started trying to target each state individually, but found it extremely difficult to find the right combination of targets for each state. After talking with @MattHJensen , we decided it would be better to just target aggregate numbers for now because we’ll be starting a project to target state levels in a couple of months and so the extra effort would be better saved until then.

I plugged in the new weights to Tax-Calculator and found aggregate tax liabilities were significantly different from what we get using our current weights file. Specifically, payroll taxes changed between 0.08 and 3.68 percent, income tax liability changed between 5.21 and 11.39 percent, and combined liabilities changed between 1.46 and 7.62 percent. For a more detailed breakdown, please see [this notebook](http://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/newcpswts.ipynb). I might add some more information to the notebook about specific variables (their totals, distributions, etc). If there is anything particular you’d like to see let me know.

edit: I've changed the last paragraph after discovering a bug in my notebook that produced incorrect tax liability calculations. The original text can be found in this comment's edit history.

cc @Amy-Xu @martinholmer


andersonfrailey:
I'll update the appropriate documentation in the coming days as well.

martinholmer:
@andersonfrailey said:
> **Third**, I renamed the directory `stage1` to `puf_stage1`
```
 stage1/growfactors.csv → puf_stage1/growfactors.csv
```

I'm confused by this change.  Tax-Calcualtor uses just **one** `growfactors.csv` file for **both** PUF and CPS input data.  Are the changes you're proposing in PR#200 advocating a different approach in Tax-Calculator?  If so, we need a lot of discussion of the pros and cons of what you're proposing.

@MattHJensen

MattHJensen:
Closing and reopening to test CI.

martinholmer:
The Travis-CI tests all pass on the tip of the master branch on local computers and at GitHub check-in.

@MattHJensen, thanks for activating Travis-CI for the taxdata repository.

@andersonfrailey, you need to merge the tip of the master branch into this development branch in order to get the tests working with each GitHub check-in.  And you will need to do some revisions to PR #200 to avoid the massive breakage of Tax-Calculator described in [this comment](https://github.com/open-source-economics/taxdata/pull/200#issuecomment-392284348) and in the new growfactors tests.

andersonfrailey:
Latest commit updates the path to the stage 1 growth factors.

andersonfrailey:
Thanks to @hdoupe for figuring out why the Travis was failing to create the new environment and how to fix it.

I've updated the [notebook](http://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/newcpswts.ipynb?flush_cashe=True) to display the results since I've made the updates to which growth rates are used.

martinholmer:
@andersonfrailey made this change in PR #200:
```
 name: taxdata-dev
+channels:
+- conda-forge
 dependencies:
```
Why are the `channels:` lines need for the taxdata-dev environment when they are not needed for the taxcalc-dev environment?

@hdoupe @MattHJensen

andersonfrailey:
@martinholmer, I believe, and @hdoupe please correct me if I'm wrong, it has to do with Conda channels. According to the [documentation](https://conda.io/docs/user-guide/tasks/manage-channels.html), different Conda channels have different packages available (with some overlap) and it seems like all of the packages in taxcalc-dev are available on the anaconda channel while the pulp package I've added in this PR is not. The difference can be seen in the install instructions.

[For Numpy](https://anaconda.org/anaconda/numpy)
`conda install -c anaconda numpy `

And [for PuLP](https://anaconda.org/conda-forge/pulp)
`conda install -c conda-forge pulp`

I don't know what travis defaults to, but it wasn't able to find and install the PuLP package so we had to specify conda-forge for it to know which channel to look for it in.

hdoupe:
@andersonfrailey yep, that's right. If you don't have the ospc channel set in your conda defaults (usually in ~/.condarc), then you have to install taxcalc via `conda install taxcalc -c ospc`

martinholmer:
@andersonfrailey and @hdoupe,  thanks for the explanation.
I didn't know that PuLP was available only on the conda-forge channel.

andersonfrailey:
@martinholmer would you be comfortable with me merging this and adopting the new weights in Tax-Calculator or are there more tests/aggregate results you'd like to see first?

martinholmer:
@andersonfrailey asked:
> would you be comfortable with me merging this and adopting the new weights in Tax-Calculator or are there more tests/aggregate results you'd like to see first?

Before merging, I'd like an answer to the question I asked nine days ago in [this comment](https://github.com/open-source-economics/taxdata/pull/200#issuecomment-392284348).

andersonfrailey:
@martinholmer, I initially used those new growth factors with the thought that it might be time to discuss using separate growth factors for the PUF and CPS because they each represent data for different years (2011 and 2014). I now believe that you're right in the sense that if we were to do that it would require a much more detailed conversation that is beyond the scope of this PR. 

I removed that portion of the PR in commit 254798d.

martinholmer:
@andersonfrailey said:
> I initially used those new growth factors with the thought that it might be time to discuss using separate growth factors for the PUF and CPS because they each represent data for different years (2011 and 2014). 

Maybe I'm missing something, but the phrase "using separate growth factors [that is, `growfactors.csv` files] for the PUF and CPS" is a logical contradiction.  As I said before, there is just one CBO projection and so there is just one `growfactors.csv` file.  We were in agreement on this fact years ago when we setup the directory structure of the taxdata repo.  Why do you think we can change it now?

> I now believe that you're right in the sense that if we were to do that it would require a much more detailed conversation that is beyond the scope of this PR.  I removed that portion of the PR in commit 254798d.

Whatever you did in that commit is not nearly enough.  We need **one** `taxdata/stage1` directory.
Can you make the changes in PR #200 to get us back to that state?  Or, if you see problems in doing that, then we need to have a phone conversation.

andersonfrailey:
@martinholmer there is still only one CBO projection. The change I initially made in this PR simply changed what we used as a start year when calculating the stage 1 factors used in stage 2 extrapolation. For any given CBO projection, the calculation for the growth rate for that year went from `projection / 2011 projection` to `projection / 2014 projection`. I've since removed that change and now there is no change in growth factors.

This PR still introduces separate stage 1 folders because the targets used by the linear programming model when creating the CPS weights are slightly different than those used to create PUF weights. As I mentioned when I opened this PR, I changed the AGI bins used for wage targeting, made the interest income target include both taxable and non-taxable interest income, and dropped a few of the other targets that were preventing the model from being solvable. All of these targets are based on IRS SOI data, not CBO projections.

As the PR stands now, the same growth factors were used to create the PUF and CPS weights files.

andersonfrailey:
Quick update. There were some code errors in the notebook I posted earlier. Specially I didn't properly calculate the change in weights or tax liability. I've updated the [notebook](http://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/newcpswts.ipynb) to use the proper code.

We now see much bigger changes in aggregate tax liability, but also more changes in the weights to account for the changes. The resulting liabilities are closer to the PUF than what we currently get with the PUF. I added an extra graph that includes the PUF totals, but it's not rendering in the notebook I've linked to for some reason. I'll try and get it updated so y'all can compare.

cc @martinholmer

martinholmer:
@andersonfrailey said in the first comment in PR #200:
>  I plugged in the new weights to Tax-Calculator and found aggregate tax liabilities to be similar to what we get using our current weights file. Specifically, payroll taxes were almost identical, income tax liability never changed more than 3.27 percent, and combined liabilities never changed more than 1.59 percent. For a more detailed breakdown, please see this [notebook](http://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/newcpswts.ipynb).

Don't you think this text need revision now that you've corrected the notebook?

martinholmer:
@andersonfrailey, Why exactly are we not using PuLP to solve for the 2014 weights?
If you look at the latest version of your notebook, you see that 2015 ITAX is up 11.4% while 2014 ITAX is essentially unchanged.  This doesn't seem plausible to me.  Why not estimate the 2014 weights using 2014 targets using exactly the same methods as for 2015+?

andersonfrailey:
@martinholmer asked:
>Don't you think this text needs revision now that you've corrected the notebook?

I'll update the comment for clarity.

>Why exactly are we not using PuLP to solve for the 2014 weights?

This is just following what we've done in the past. For both the CPS and the PUF we've used the weights on the original file for that first year rather than using an LP model to adjust them.

martinholmer:
@andersonfrailey said:
>> Why exactly are we not using PuLP to solve for the 2014 weights?
>
> This is just following what we've done in the past. For both the CPS and the PUF we've used the weights on the original file for that first year rather than using an LP model to adjust them.

But for the CPS file we have "2014" weights from three years and, I'm assuming, we have little documentation about how/if the weights from three different CPS surveys have been made consistent.

Doesn't the enormous difference between the PuLP-derived 2015 weights and the raw 2014 weights make you nervous?  Why don't you use PuLP to compute 2014 weights and see if things look more sensible when comparing tax revenues in 2014 and 2015?

andersonfrailey:
@martinholmer I'll run the model to include the 2014 weights and see what we get. I'll post results here when I finish.

andersonfrailey:
Re-ran the LP model also solving for 2014 weights. This resulted in 2014 tax liability dropping about 5%. Full results can be seen in this [notebook](http://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/newcpswts.ipynb).

andersonfrailey:
@martinholmer I haven't pushed the changes yet. Was planning on waiting until we decided if this was the route we wanted to go down. All that's changed is I called the `solve_lp_for_year` function for 2014 and I dropped two targets: dividends and unemployment compensation. This was because the model was infeasible at all reasonable tolerances when including them.

andersonfrailey:
@martinholmer the LP model can't solve for the 2014 weights when including dividend and UI targets because the aggregate totals in the file are simply too far from the target levels for those variables. As @Amy-Xu can attest there's only so much the LP model can make up for.

I'll just push the changes that would result in 2014 weights here. It'll be easier for me at least to revert those changes that keep up with another branch and PR.

andersonfrailey:
@martinholmer it has to do with our targets for unemployment. In 2014, we set an unemployment target of $33,288,898, in accordance with unemployment compensation reported to the IRS for that year. That can be found in table 1.4 [here](https://www.irs.gov/statistics/soi-tax-stats-individual-statistical-tables-by-size-of-adjusted-gross-income). Our growth rate for unemployment compensation for 2015 is 0.3603, which results in a 2015 target of $10,925,416.32.

There is also a relatively large change in the dividend target between the years, going from $254,702,232.00 to $388,777,486.93 (all of the targets are in thousands of dollars). These large changes in targets between the years would explain why the model is infeasible in 2014 when including them, but feasible in 2015.

martinholmer:
@andersonfrailey said in PR #200:
> Using the LP model for 2014 resulted in a new minimum weight (149 vs 150) so I updated the weight test accordingly. I also noticed the error message for that test printed the max weight rather than the minimum weight so I updated that too.

Looks good to merge #200.  Thanks for fixing my mistake in the weights test.

PR #211, opened 2018-05-29
Add taxdata/puf_stage2/stage2.sh bash shell script

martinholmer:
This pull request adds a shell script to the `puf_stage2` directory that stops the basic `stage2.py` program when the LP solver writes a message indicating that solving the weights LP problem is infeasible with the specified LP tolerance.  If no such message is written by the LP solver, then the basic `stage2.py` program runs to completion.  In either case, all the screen output from the `stage2.py` program is saved to a file named `stage2.log`.   When  the basic `stage2.py` program runs to completion, the log file can be copied to a file that is under version control, generating a history of successful LP solver results.  The log copy for the current version of the taxdata repository, which is included in this pull request, is called `stage2.results`.

When an infeasibility is encountered, the `stage2.sh` output looks like this:
```
iMac2:puf_stage2 mrh$ ./stage2.sh 
Tue May 29 15:49:59 EDT 2018
INFEASIBLE STOP: read stage2.log file for details
Tue May 29 15:50:19 EDT 2018
```
And when no infeasibility message is written, the output looks like this:
```
iMac2:puf_stage2 mrh$ ./stage2.sh 
Tue May 29 18:16:04 EDT 2018
FINISHED EXECUTION: read stage2.log file for details
Tue May 29 18:47:37 EDT 2018
```

This new `stage2.sh` script is meant to replace the notebook approach to catching infeasibility, which requires the time-consuming and error-prone process of maintaining duplicate code in the notebook and in the `stage2.py` program.

@andersonfrailey, does this capability meet your needs?
If so, the same script (perhaps with minor edits) can be used in the `taxdata/cps_stage2` directory.


martinholmer:
@andersonfrailey said:
> I've been doing a little research on ways to monitor output in python and I believe we can add a function to `stage2.py` that runs in parallel to the `solve_lp_for_year` function, periodically checking its output, like you've done with the shell script in this PR, and if it detects an infeasibility, stops `solve_lp_for_year` and then calls the function again with a higher tolerance. This would automate steps 2 and 3 in my workflow and save the user from having to start the process from the beginning each time they encounter a failure.

Yes, now that you actually explain what you do by hand, the idea you describe above would provide more automation because it automatically revises the tolerance values in the `stage2.py` program.

While you work on that more sophisticated capability, I'll merge pull request #211 211.

PR #215, opened 2018-05-31
Add variable value and presence tests

andersonfrailey:
This PR adds a test that checks the unweighted aggregate values in `cps.csv.gz` and `puf.csv`

It should fail each time either file is changed, which allows the used to make sure only the variables they're modifying are changed.

I'll be modifying this PR so that a different message is shown when an unexpected variable is included in the file. That commit should be coming shortly.

cc @MattHJensen


andersonfrailey:
Latest commit adds a check for new and missing variables. I've changed the test name to `variable_check` because that seems more appropriate now that it's going beyond just aggregate values tests.

MattHJensen:
What's your thinking on weighted vs unweighted?

andersonfrailey:
@MattHJensen I think a test that includes weights is a good idea. Do you think just looking at the sum of the weights for each year would be sufficient or should we go all out and find weighted sums for each year?

martinholmer:
There’s already a test that sums the weights for each year and checks that the sum is within a fairly narrow range. The existing test is strong enough to catch the kind of error that occurred about a week ago (when incorrect weights made income tax code

andersonfrailey:
@MattHJensen, given that we already have a set of tests for the weights, I don't think a weighted sum check is really necessary right now. I might change my mind and come back to it at a later date though.

MattHJensen:
Ok, thanks for the information @andersonfrailey and @martinholmer!

andersonfrailey:
Great. I'm going to merge this PR and we'll circle back to creating tests for aggregate totals.

PR #229, opened 2018-06-19
New LP Model for PUF

andersonfrailey:
This PR introduces a new linear programming model for creating the `puf_weights.csv.gz` file. This moves us fully away from the CyLP model, which was found to be difficult to install and forced us to pin versions of Scipy and Numpy in the taxdata environment.

The resulting aggregate tax liabilities are virtually unchanged from current results. The largest change is a 1.13% drop in income tax liabilities in 2016. Other than that all changes are less than +/- one percent. Full results can be seen [here](http://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/PUF_PuLP_Wts.ipynb?flush_cashe=True).

cc @martinholmer @MattHJensen


martinholmer:
@andersonfrailey, PR #229 looks good to me.
Thanks for getting the repo moved to a more current LP solver.

PR #230, opened 2018-06-22
Update dependencies

lucassz:
Update dependencies as requested by @andersonfrailey. Everything seems to work fine with this change.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,9 +3,10 @@ channels:
 - conda-forge
 dependencies:
 - python>=2.7.14
-- numpy=1.12.1
-- pandas=0.20.3
-- scipy=0.18.1
+- numpy>=1.12.1
+- pandas>=0.20.3
+- scipy>=0.18.1
 - bokeh >=0.12.3
 - pytest
 - pulp
+- tqdm


andersonfrailey:
Thanks, @lucassz. This looks good to me. I'll merge at the end of the day if there are no more comments.

martinholmer:
@lucassz revised the `environment.yml` file to include this:
```
python>=3.6.3
```
Why can't developers use Python 2.7?

@andersonfrailey

andersonfrailey:
@martinholmer, the biggest reason in my opinion is Pandas will be dropping Python 2.7 support at the end of the year and I believe TaxData would be best off if we've moved to Python 3.6 before that happens. TaxData is currently completely compatible with Python 3.6 so I see no reason why we should spend the next six months worrying about 2.7 compatibility when we can make the switch now and avoid any potential issues.

If a user is working with TaxData, they're running the scripts in the TaxData environment, not integrating it with other packages so breaking backwards compatibility with other projects isn't an issue. This is different from Tax-Calculator, which needs to maintain compatibility with projects like PolicyBrain and BRC.

Making the move to 3.6 will also allow us to use new features such as [f-strings](https://cito.github.io/blog/f-strings/) in the repo. Replacing `.format()` with an f-string may seem trivial, but it makes code much cleaner and readable.

martinholmer:
@andersonfrailey,  Can we wait on making taxdata an only-3.6 project until **all** the taxdata developers have moved from Python 2.7 to Python 3.6?

andersonfrailey:
@martinholmer, we could, but I don't see any advantages to waiting. If a developer is setting up their environment using the `environment.yml` file then beyond having to update the taxdata-dev environment their work flow won't be changing in any significant way.

martinholmer:
@andersonfrailey said:
>> Can we wait on making taxdata an only-3.6 project until **all** the taxdata developers have moved from Python 2.7 to Python 3.6?
>
> We could, but I don't see any advantages to waiting. If a developer is setting up their environment using the `environment.yml` file then beyond having to update the `taxdata-dev` environment their work flow won't be changing in any significant way.

Can we talk about this on Monday in DC?

andersonfrailey:
@lucassz, @martinholmer and I talked yesterday about why he would like us to delay the move to python 3.6 and having heard his reasoning I think it makes sense. Can you update this PR to undo the Python version change? After that I'll go ahead and merge it. We'll make the move to python 3.6 in the coming months.

lucassz:
@andersonfrailey Yes, I just amended the commit.

andersonfrailey:
Thanks, @lucassz. Merging.

PR #238, opened 2018-07-02
Update taxdata/README.md info and csvcopy.sh script

martinholmer:
This pull request updates the main `README.md` file to reflect the use of the `taxdata/Makefile` as the way to do data-preparation in the taxdata repository.

It also updates the `csvcopy.sh` script to reflect the fact that the `growfactors.csv` file has recently been moved from the `stage1` directory to the `puf_stage1` directory.

@andersonfrailey and @hdoupe, can you suggest improvements in the new documentation?  Do you have questions that the new documentation does not answer?

With respect to the old plan of having a `README.md` file in each subdirectory, it seems as if that approach to documentation has been far from ideal.  What about just removing all of that documentation because that information is in the `Makefile`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,22 +44,28 @@ The best documentation of the data-preparation workflow is the
 the Python script that generates each made file.  The files made in
 early stages of the workflow serve as input files in later stages,
 which means there is a cascading effect of changes in the scripts
-and/or input files.  The Makefile handles all this complexity in an
-economical way because it executes scripts to make new versions of
+and/or input files.  The Makefile automates this complex workflow in
+an economical way because it executes scripts to make new versions of
 made files only when necessary.  Start exploring the Makefile by
-running the `make help` command.
+running the `make help` command.  If you want more background on the
+make utility and makefiles, search for Internet links with the
+keywords `makefile` and `automate`.
 
 Note that the stage2 linear program that generates the weights file is
 very long-running, taking five or more hours depending on your
 computer's CPU speed.  We are considering options for speeding up this
 stage2 work, but for the time being you can execute `make puf-files`
 and `make cps-files` in separate terminal windows to have the two
-stage2 linear programs run in parallel.  If you are generating the
+stage2 linear programs run in parallel.  (If you try this parallel
+execution approach, be sure to wait for the `make puf-files` job to
+begin stage2 work before executing the `make cps-files` command in
+the other terminal window.  This is necessary because the CPS stage1
+work depends on output from PUF stage1.)  If you are generating the
 taxdata made files in an overnight run, then simply execute the `make
 all` command.
 
 You can copy the made files to your local Tax-Calculator directory
-tree using the [`csvcopy.sh` bash script](csvcopy.sh).  Use the dryrun
+tree using the [`csvcopy.sh`](csvcopy.sh) bash script.  Use the `dryrun`
 option to see which files would be copied (because they are newer than
 the corresponding files in the Tax-Calculator directory tree) without
 actually doing the file copies.  At the terminal command-prompt in the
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,88 +1,86 @@
...
+running the `make help` command.
+
+Note that the stage2 linear program that generates the weights file is
+very long-running, taking five or more hours depending on your
+computer's CPU speed.  We are considering options for speeding up this
+stage2 work, but for the time being you can execute `make puf-files`
+and `make cps-files` in separate terminal windows to have the two
+stage2 linear programs run in parallel.  If you try this, be sure to
+wait a bit until the `make puf-files` starts on stage2 before running
+the `make cps-files` command in the other terminal window.  If you are
```
andersonfrailey: Out of curiosity, is there a reason they should wait before running `make cps-files`? Does it have something to do with how makefiles work or did you run into a problem running `make puf-files` and `make cps-files` at the same time?

martinholmer: It's because `make cps-files` and `make puf-files` both generate the intermediate factors files.
It seems as if running them simultaneously might cause confusion with the date-stamp on the made files.  Doing one first (which takes just a moment because stage1 is fast) means that when you start the second, the Makefile will skip stage1.  Does this make sense?

andersonfrailey: Yes, that makes sense. Thanks for the clarification.

martinholmer:
@hdoupe, Earlier you had made a request for documentation on the make utility and makefiles and how they are being used in the taxdata repository.  PR #238 in the taxdata repo adds some documentation.  Is this new documentation sufficient to answer your questions?  If not, what additional topics should the documentation cover?

hdoupe:
This looks good to me. Thanks for asking for my input.

I didn't realize that `Makefile`s only create files that are not there. Does `make` only re-create the file if it is not there or if one of the input files is not there? For example, in the below target, the `cps_stage_1/stage_2_targets.csv` would be re-created if `cps_stage1/stage1.py` was changed? 


```
cps_stage1/stage_2_targets.csv: cps_stage1/stage1.py \
                                cps_stage1/SOI_estimates.csv \
                                puf_stage1/Stage_I_factors.csv \
                                puf_stage1/Stage_II_targets.csv
	cd cps_stage1 ; python stage1.py
```

martinholmer:
@hdoupe asked:
> I didn't realize that Makefiles only create files that are not there. 

Actually, they run the recipe for the target only if the target is out-of-date, meaning that one or more of the dependencies has a newer time-stamp than the target file's time-stamp.  So, for example,
```
target: dependency
    recipe
```
means that the recipe executes only if the dependency is newer than the target.

@hdoupe continued:
> Does make only re-create the file if it is not there or if one of the input files is not there? For example, in the below target, the `cps_stage_1/stage_2_targets.csv` would be re-created if `cps_stage1/stage1.py` was changed?
```
cps_stage1/stage_2_targets.csv: cps_stage1/stage1.py \
                                cps_stage1/SOI_estimates.csv \
                                puf_stage1/Stage_I_factors.csv \
                                puf_stage1/Stage_II_targets.csv
	cd cps_stage1 ; python stage1.py
```
Answer to your first question is NO.
Answer to your second question is YES because changing the `cps_stage1/stage1.py` file would mean its time-stamp was more recent than the time-stamp on the target `cps_stage_1/stage_2_targets.csv` file.

Of course, the recipe would execute for any target file that did not exist, but the make utility is more nuanced than that.

Does all this make sense?

hdoupe:
Yep, thanks for the explanation. make seems like a pretty neat tool. I think most people these days aren't exposed to it since it seems more associated with compiled languages (i.e. compiling and installing packages). This association also makes it seem scarier then it actually is.

Anyways, I've enjoyed using it in PolicyBrain, too.

martinholmer:
@andersonfrailey and @hdoupe,
Given the discussion of PR #238, I plan to merge it later today if you have no objections.

PR #240, opened 2018-07-05
Add re-zip logic to taxdata Makefile

martinholmer:
No change in data or logic; just set time-stamp in each GZIP file to zero so that `git` accurately reports the status of the contents of each `.csv.g`z file.  If this is not done, `git` will report that, for example, `cps_weights.csv.gz` is different even though the `cps_weights.csv` contents are exactly the same as before.  `git` will think this because the `.csv.gz` time-stamp changes each time the `.csv.gz` file is created.  There is no option in the pandas `to_csv` function to control the GZIP file header's time-stamp value; hence the use of the `gzip` utility in the new `Makefile`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,8 @@
 #              command-line tools installed), so this Makefile is designed
 #              to work in that environment (and not on Windows).
 # Note that the re-zipping of each .csv.gz file "removes" the time-stamp
-#              from the GZIP file header, which is necessary for correct
-#              operation of this Makefile.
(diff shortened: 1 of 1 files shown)


andersonfrailey:
This looks good, @martinholmer. Thanks for figuring out how time-stamps affect git behavior with these files.

martinholmer:
@andersonfrailey,  Thanks for the quick review. 
I'll merge this now so you can use the new Makefile in your final preparations of PR #185.

Issue #241, opened 2018-07-05
Questions about extrapolated benefits


martinholmer:
Given the fact that (1) the `cps_stage4/extrapolation.py` script on the taxdata master branch does not work, and (2) the `taxcalc/cps_benefits.csv.gz` file on the Tax-Calculator master branch was apparently generated with unknown weights, and (3) the `test_extrapolation.py` tests were never finished, I decided it would be prudent to compare the actual benefit dollar amounts and benefit recipient counts generated by Tax-Calculator with the dollar and count targets implied by the `cps_stage4/growth_rates.csv` data (which is used by the `cps_stage4/extrapolation.py` script.

I had difficulty doing this because when I looked in the C-TAM and taxdata repositories, I could not find the benefit dollar and recipient count targets for each benefit program by year.  Did I miss that crucial information?

Given that I couldn't find that information, I developed a script to generate for each program year two targets: one for aggregate dollar benefits and the other for total number of participants or recipients.  That script, which is called `bentarget.py`, tabulates the actual benefit dollars and recipient counts for 2014 using Tax-Calculator.  And then those base-year targets are inflated to the comparison year using the appropriate growth rate from the `taxdata/cps_stage4/growth_rates.csv` file.  Then in another script, which is called `benactual.py`, Tax-Calculator is used to tabulate the aggregate dollar benefits and  total number of filing unit recipients for the comparison year.  These two scripts write out the constructed targets and the actual Tax-Calculator results for both the base year (2014) and the comparison year (either 2015 or 2027) to files that have the same format.  And finally, those two target and actual results files are compared with the Unix diff utility.  I will post these two scripts in a subsequent comment in this issue.

Here I show the results of this exercise and then make some observations and ask some questions.
In all the results tables shown below, the third column is aggregate benefits in billions of dollars, the fourth column is number of recipients in millions of filing units, and the fifth column the the ratio of the two, which is the average benefit expressed in thousands of dollars per year.

**2014 RESULTS**
```
2014 ssi	  54.040   6.829     7.9
2014 mcare	 575.245  38.534    14.9
2014 mcaid	 368.102  27.905    13.2
2014 snap	  82.909  28.516     2.9
2014 wic	   3.602   4.817     0.7
2014 tanf	  30.891   3.388     9.1
2014 vet	 146.602   4.901    29.9
2014 housing	  32.604   4.626     7.0
```

**2015 RESULTS with `<` denoting target results and `>` actual results**
```
[...]

Issue #232, opened 2018-06-22
Benefits extrapolation script does not work


martinholmer:
On the tip of the master branch, I ran the `rename_columns.py` script successfully.
Then I changed the `extrapolation.py` script to read the CPS weights in the `cps_stage2/cps_weights.csv.gz` file and when I execute `python extrapolation.py` this is what I get:
```
iMac:cps_stage4 mrh$ python extrapolation.py 
('starting year', 2015)
extrapolation.py:148: FutureWarning: Sorting because non-concatenation axis is not aligned. A future version
of pandas will change to not sort by default.

To accept the future behavior, pass 'sort=True'.

To retain the current behavior and silence the warning, pass sort=False

  result = pd.concat([noncandidates, candidates], axis=0, ignore_index=False)
('total time', 6.2189788818359375)
('ssi', 'this year total is ', 7526933, ', target is ', '7537656.678923858', ', diff is ', '-10723.67892385833', ',', '-0.0014226807323080859')
('total time', 4.664236068725586)
('mcaid', 'this year total is ', 56293352, ', target is ', '56364501.59841771', ', diff is ', '-71149.59841770679', ',', '-0.0012623122071517463')
('total time', 4.444383144378662)
('mcare', 'this year total is ', 50637924, ', target is ', '50662439.87755156', ', diff is ', '-24515.877551563084', ',', '-0.0004839063734556935')
('total time', 4.628061771392822)
('vet', 'this year total is ', 4917104, ', target is ', '4939167.070071559', ', diff is ', '-22063.0700715594', ',', '-0.004466961687781042')
('total time', 4.7146289348602295)
('snap', 'this year total is ', 42180559, ', target is ', '42189411.44151255', ', diff is ', '-8852.441512547433', ',', '-0.0002098261438138247')
('total time', 4.91064715385437)
('tanf', 'this year total is ', 2979556, ', target is ', '3001327.9533102163', ', diff is ', '-21771.95331021631', ',', '-0.0072541067317230855')
Traceback (most recent call last):
  File "extrapolation.py", line 264, in <module>
    ben.increment_year()
  File "extrapolation.py", line 45, in increment_year
    tol=tol)
  File "extrapolation.py", line 133, in _extrapolate
    atol=0.0, rtol=tol
AssertionError
```
What's going on?

martinholmer:
@hdoupe and @andersonfrailey,  When was the last time somebody executed the `extrapolation.py` script?

[...]

PR #242, opened 2018-07-05
Revise cps_stage4/extrapolation.py script

martinholmer:
This pull request makes changes that allow the `extrapolation.py` script to run to completion and to pass the taxdata repo's code-style test.  The first commit in this PR increases the relative tolerance (for an acceptable extrapolation) from 0.01 to 0.05, which allows the script to run to completion.  The second commit eliminates several pycodestyle (nee PEP8) [warnings](https://github.com/open-source-economics/taxdata/issues/232#issuecomment-399711170) such as line too long, too many spaces, and ambiguous variable name.  The changes in the second commit in this PR produce exactly the same `cps_benefits.csv.gz` file as the first commit produced, confirming that the changes in the second commit are purely cosmetic.

However, the contents of the new `cps_benefits.csv.gz` file produced with this code are substantially different from the contents of the `cps_benefits.csv.gz` file on the taxdata master branch (which is currently being used in Tax-Calculator).  The next step is to determine if this new `cps_benefits.csv.gz` file has sensible content.  We plan to do that by using it (and other more current taxdata made files) in the kind of "test" described in taxdata issue #241.

@hdoupe and @andersonfrailey, I'd appreciate it if you could review these changes.

Also, @hdoupe, I need your direction on how to handle the pandas warning described at the bottom of [this 232 comment](https://github.com/open-source-economics/taxdata/issues/232#issuecomment-399711170).

### DIFF
--- a/cps_stage4/extrapolation.py
+++ b/cps_stage4/extrapolation.py
@@ -24,7 +24,7 @@ class Benefits():
         self.benefit_names = benefit_names
 
 
-    def increment_year(self, tol=0.01):
+    def increment_year(self, tol):
         self.current_year += 1
         print("starting year", self.current_year)
         WT = self.WT.loc[:, 'WT'+str(self.current_year)]
@@ -79,7 +79,7 @@ class Benefits():
 
     @staticmethod
     def _extrapolate(WT, I, benefits, prob, target,
-                     benefit_name, benefit_year, tol=0.01, J=15):
+                     benefit_name, benefit_year, tol, J=15):
         """
         Goal: get number of participants as close to target as possible
         Steps:
@@ -283,7 +283,7 @@ if __name__ == "__main__":
     ben = Benefits()
 
     for _ in range(13):
-        ben.increment_year()
+        ben.increment_year(tol=0.05)
 
     # drop unnecessary variables
     drop_list = []
--- a/cps_stage4/extrapolation.py
+++ b/cps_stage4/extrapolation.py
@@ -65,11 +65,11 @@ class Benefits(object):
             setattr(self, '{}_benefits'.format(bname), benefits)
 
     @staticmethod
-    def extrapolate(WT, III, benefits, prob, target,
+    def extrapolate(WT, part, benefits, prob, target,
                     benefit_name, benefit_year, tol, maxsize=15):
         """
         Arguments:
-          - III is boolean array of benefit program participation.
+          - part is boolean array indicating benefit program participation.
           - maxsize is the assumed maximum number of people in a filing unit.
         Goal: get number of participants as close to target as possible
         Steps:
@@ -96,37 +96,37 @@ class Benefits(object):
         """
         start = time.time()
 
-        extrap_df = Benefits._stack_df(WT, III, benefits, prob, maxsize)
+        extrap_df = Benefits._stack_df(WT, part, benefits, prob, maxsize)
 
-        receives = extrap_df.loc[extrap_df.III > 0, ]
-        avg_benefit = receives.benefits_wt.sum() / receives.III_wt.sum()
+        receives = extrap_df.loc[extrap_df.part > 0, ]
+        avg_benefit = receives.benefits_wt.sum() / receives.part_wt.sum()
 
-        actual = extrap_df.III_wt.sum()
+        actual = extrap_df.part_wt.sum()
         diff = actual - target
         if diff < 0:
             remove = False
-            candidates = extrap_df.loc[extrap_df.III == 0, ].copy()
-            assert candidates.III.sum() == 0
+            candidates = extrap_df.loc[extrap_df.part == 0, ].copy()
+            assert candidates.part.sum() == 0
             # set everyone as a participant and then remove surplus below
-            # this is necessary since III * wt will all be zeroes
-            candidates.loc[:, 'III'] = np.ones(len(candidates))
-            candidates.loc[:, 'III_wt'] = candidates.III * candidates.wt
-            noncandidates = extrap_df.loc[extrap_df.III == 1, ].copy()
-            assert noncandidates.III.sum() == len(noncandidates)
+            #   this is necessary since part * wt will all be zeroes
+            candidates.loc[:, 'part'] = np.ones(len(candidates))
+            candidates.loc[:, 'part_wt'] = candidates.part * candidates.wt
+            noncandidates = extrap_df.loc[extrap_df.part == 1, ].copy()
+            assert noncandidates.part.sum() == len(noncandidates)
         else:
             remove = True
-            candidates = extrap_df.loc[extrap_df.III == 1, ].copy()
-            assert candidates.III.sum() == len(candidates)
-            noncandidates = extrap_df.loc[extrap_df.III == 0, ].copy()
-            assert noncandidates.III.sum() == 0
-        noncan_part = noncandidates.III_wt.sum()
+            candidates = extrap_df.loc[extrap_df.part == 1, ].copy()
+            assert candidates.part.sum() == len(candidates)
+            noncandidates = extrap_df.loc[extrap_df.part == 0, ].copy()
+            assert noncandidates.part.sum() == 0
+        noncan_part = noncandidates.part_wt.sum()
         del extrap_df
         # sort by probability of getting benefits in descending order
-        # of prob so that we only keep those with highest prob of
-        # receiving benefits in the future
+        #   of prob so that we only keep those with highest prob of
+        #   receiving benefits in the future
         candidates.sort_values(by='prob', ascending=False, inplace=True)
         # create running sum of weighted participants
-        candidates['cum_participants'] = (candidates.III_wt.cumsum() +
+        candidates['cum_participants'] = (candidates.part_wt.cumsum() +
                                           noncan_part)
         # get absolute difference between candidates and target
         candidates['_diff'] = candidates.cum_participants - target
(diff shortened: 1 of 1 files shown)


cps_stage4/extrapolation.py
```diff
@@ -78,8 +83,8 @@ def increment_year(self, tol=0.01):
             setattr(self, '{}_benefits'.format(benefit), benefits)
 
     @staticmethod
-    def _extrapolate(WT, I, benefits, prob, target,
-                     benefit_name, benefit_year, tol=0.01, J=15):
+    def _extrapolate(WT, III, benefits, prob, target,
+                     benefit_name, benefit_year, tol, JJJ=15):
```
hdoupe: @martinholmer why did you change the variable names here?

cps_stage4/extrapolation.py
```diff
@@ -205,85 +212,92 @@ def _read_data(self, growth_rates, cps_benefit, cps_weights,
...
                                  cps_benefit.WT).sum()
             participant_targets = total_particpants * \
-                (1 + growth_rates['{0}_participation_growth'.format(benefit)])
-            # extract probability of participation in program from tax-unit database
+                (1 + growth_rates['{}_participation_growth'.format(benefit)])
+            # extract probability of participation in program from
+            # tax-unit database
             prob_col = [col for col in list(cps_benefit)
-                        if col.startswith('{0}_PROB'.format(benefit.upper()))]
+                        if col.startswith('{}_PROB'.format(benefit.upper()))]
```
hdoupe: @martinholmer what's the benefit of using `'{}'.format(...)` over `'{0}'.format(...)`

hdoupe:
@martinholmer said:

> However, the contents of the new `cps_benefits.csv.gz` file produced with this code are substantially different from the contents of the `cps_benefits.csv.gz` file on the taxdata master branch (which is currently being used in Tax-Calculator). The next step is to determine if this new `cps_benefits.csv.gz` file has sensible content. We plan to do that by using it (and other more current taxdata made files) in the kind of "test" described in taxdata issue #241.

Is there a way that we can examine how the data that is used as an input to this script has changed over time? Perhaps, something has changed in this script, but if the inputs data is different then, that would explain why the tolerance needed to be raised and why the results are so different.

hdoupe:
@martinholmer said:
> Also, @hdoupe, I need your direction on how to handle the pandas warning described at the bottom of this [232 comment](https://github.com/open-source-economics/taxdata/issues/232#issuecomment-399711170).

I'm looking into this now.

cps_stage4/extrapolation.py
martinholmer: It's one character less.

cps_stage4/extrapolation.py
martinholmer: To avoid the pucodestyle warnings about ambiguous variable name.
See [this list](https://github.com/open-source-economics/taxdata/issues/232#issuecomment-399711170) of warnings.

martinholmer:
@hdoupe said:
> if the input data are different, then that would explain why the tolerance needed to be raised and why the results are so different.

Yes, I'm sure the inputs have changed.  The `extrapolation.py` script hasn't been run in months and when I tried to run it it crashed because it was reading in a file that did not exist in the taxdata repository.  I think things were way too messed up to do a backward-looking investigation.  My focus is on looking forward and getting sensible (that is, tested) benefits extrapolation results.  I think the preliminary testing work in issue #241 already shows previously undetected problems with the extrapolated TANF data.

hdoupe:
@martinholmer said:
> My focus is on looking forward and getting sensible (that is, tested) benefits extrapolation results. I think the preliminary testing work in issue #241 already shows previously undetected problems with the extrapolated TANF data.

OK, that sounds good to me.

cps_stage4/extrapolation.py
hdoupe: OK, thanks for the explanation.

cps_stage4/extrapolation.py
hdoupe: I see, but `III` doesn't seem to add any more information about what the matrix is than the information given by `I`. Could we use something like `indicator` or `ind`, given that this is an indicator matrix. `JJJ` could be `num_members_in_unit` or something else like that. Does that seem sensible?

hdoupe:
The line of code that causes the pandas warning is:

```
result = pd.concat([noncandidates, candidates], axis=0, ignore_index=False)
```

This is taking two dataframes:
- `noncandidates` -- a `N_n*J` by `4` matrix (why 4?: see columns bullet point)
  - `N_n` the number of people who are not candidates for a change in participation status 
  - `J` - max number of people in each unit (15 for our data set)
  - Columns: 'wt', 'I', 'benefits', 'prob' (weight, participation status indicator, benefit amount received, probability of participating in program)
- `candidates` -- a `N_c*J` by `4`
  - `N_c` the number of people who are candidates for a change in participation status 
  - same definitions for other dimensions

This line of code produces one dataframe:
- `results` -- a `J * (N_n + N_c)` by `4` matrix

Thus, the two dataframes are simply stacked on top of each other. They are aligned by their "columns". That is the `I` column of `candidates` is matched to the `I` column of `noncandidates` and so on with the other columns. Both dataframes have the exact same columns and these columns have the exact same order.

Pandas issue pandas-dev/pandas#4588 discusses whether pandas should sort these columns (or rows if `axis=1`) by default if they are not already aligned. If you pass `sort=True`, then pandas will sort both dataframes by their columns (or rows) using alphanumeric ordering. If you pass `sort=False`, then pandas will not sort the dataframes no matter whether the columns (or rows) are aligned or not.

In our case, both dataframes ought to have the same columns. I am unfamiliar with the inner workings of pandas and thus, cannot say whether the columns are guaranteed to have the same ordering or not. So, my recommendation is to pass `sort=True`. `join` can be set to `inner` or `outer`. Both dataframes should have the same columns--no more and no less. So, it doesn't matter whether the join is `inner` or `outer`.

This explanation is based on my reading of the docs. If anyone has a different interpretation of them, then I'm interested in hearing it. Also, please let me know if this explanation is unclear or could be improved. I hope this explanation clears things up.

@martinholmer thanks for laying out the problem clearly and directing me to the relevant portion of the pandas docs.

cps_stage4/extrapolation.py
martinholmer: @hdoupe said:
> I see, but `III` doesn't seem to add any more information about what the matrix is than the information given by `I`. 

Yes, you are absolutely right about that.

> Could we use something like `indicator` or `ind`, given that this is an indicator matrix. `JJJ` could be `num_members_in_unit` or something else like that. Does that seem sensible?

I like your proposed rename of `JJJ` because the variable name says what it is.
But changing `III` to `indicator` still leaves readers of the script wondering what is being indicated.
What does `III` indicate?

hdoupe: Ah, right. By "indicator" I mean: indicator of whether the person is participating or not. How about `participating_switch` or `participating_indicator`? These are still pretty long though.

martinholmer: @hdoupe said:
> `JJJ` could be `num_members_in_unit` or something else like that.
Do you mean number of people in a filing unit?  If so, that doesn't seem right given the code.
For example, what does `JJJ=15` mean?  Can you clarify?

hdoupe: My understanding is that `J=15` means that there are at a maximum 15 members in each unit. That means that the maximum width of the participation indicator matrix is 15.

martinholmer:
@hdoupe, We need to revisit the issue of the Pandas warning.

The warning (the text of which I post below) arises when this line is executed:
```
        result = pd.concat([noncandidates, candidates], axis=0,
                           ignore_index=False)
```
So, you are merging the `noncandates` and the `canditates` dataframes on `axis=0`, which is the rows.
Is that what you want to do?  

The complete warning is this:
```
extrapolation.py:159: FutureWarning: Sorting because non-concatenation axis is not aligned. A future version
of pandas will change to not sort by default.

To accept the future behavior, pass 'sort=True'.

To retain the current behavior and silence the warning, pass sort=False
  ignore_index=False)
```
I have no idea what you're doing here, so you're going to have to give me more direction.
To me, the warning says the order of the columns is not the same in the two dataframes.  Is that true or false?

hdoupe:
@martinholmer asked: 
> So, you are merging the noncandidates and the candidates dataframes on axis=0, which is the rows.
Is that correct? Is that what you want to do? I don't see how the rows match up since the names of the two dataframes suggest they are two distinct groups.

This isn't quite correct. Consider this toy example:

```
>>> import pandas as pd
>>> df = pd.DataFrame({"a": [1, 2], "b": [3, 4]})
>>> df
   a  b
0  1  3
1  2  4
>>> df2 = pd.DataFrame({"a": [5, 6], "b": [7, 8]})
>>> df2
   a  b
0  5  7
1  6  8
>>> r = pd.concat([df, df2], axis=0)
>>> r
   a  b
0  1  3
1  2  4
0  5  7
1  6  8
>>> r.a
0    1
1    2
0    5
1    6
Name: a, dtype: int64
>>> r.b
0    3
1    4
0    7
1    8
Name: b, dtype: int64
>>> 
```

There are two dataframes with the same columns "a" and "b". Both columns in both data frames have two elements in them. The dataframes are stacked on top of each other. The result is one dataframe with two columns "a" and "b" and four elements in each column. 

This is the same thing that occurs in `extrapolation.py`. Does this make sense?

> I have no idea what you're doing here, so you're going to have to give me more direction.
To me, the warning says the order of the columns is not the same in the two dataframes. Is that true or false?

I'm not sure. Both dataframes have the same columns. There could be an internal pandas bug or something in the script that swaps the order of the columns. If the column order is changed, then yes, they would need to be sorted. However, I'm not sure why the order would change.

I hope this helps. I'm happy to continue going over how this script works and working to make it easier to understand.

martinholmer:
@hdoupe,  Thanks for the detailed explanation of `pd.concat` logic in pull request #242.

I've added `sort=False` to the `pd.concat` call and the `extrapolation.py` results are unchanged.  Does that seem like the correct way to suppress the warning message?

martinholmer:
@hdoupe and @andersonfrailey, I think PR #242 is ready for review.  All the changes today are either cosmetic style changes or eliminate the Pandas concat warnings.  Today's changes produce exactly the same results as before today's changes.

andersonfrailey:
Thanks for working on this @martinholmer and @hdoupe. LGTM, but I'd defer to @hdoupe's final judgement.

hdoupe:
Yes, this looks good to me. Thanks for making these updates @martinholmer.

I'll take a look at the failing tests soon. It looks like the path names were never moved into fixtures in addition to some other maintenance work that they will need.

PR #248, opened 2018-07-12
Reduce rtol_amt and rtol_cnt in test_extrapolated_benefits

martinholmer:
This pull request reduces to the lowest possible values the `rtol_amt` and `rtol_cnt` relative tolerances for actual-vs-target comparisons for benefit dollar amounts and benefit recipient counts.  By lowest possible, we mean that lower values will cause the test to fail.  
The new lower values are `rtol_amt = 0.002` and `rtol_cnt = 0.10`

If we temporarily set `rtol_cnt = 0.05`, which is what it should be given the current version of the `cps_stage4/extrapolation.py` script, there are many test failures.  Here is a list of the test failures:
```
--------------------------------------------------
TEST RESULTS WITH rtol_amt=0.002 and rtol_cnt=0.05
--------------------------------------------------
2015 ssi	CNT	    6.448    6.815    -5.4
2015 snap	CNT	   25.292   27.968    -9.6
2016 mcaid	CNT	   29.175   30.949    -5.7
2017 mcaid	CNT	   29.674   31.505    -5.8
2018 mcaid	CNT	   30.175   32.063    -5.9
2018 tanf	CNT	    3.101    2.930     5.8
2019 mcaid	CNT	   30.680   32.576    -5.8
2019 tanf	CNT	    3.145    2.930     7.3
2019 housing	CNT	    4.236    4.473    -5.3
2020 mcaid	CNT	   31.165   33.093    -5.8
2021 mcaid	CNT	   31.638   33.519    -5.6
2021 housing	CNT	    4.146    4.407    -5.9
2022 mcaid	CNT	   32.118   33.949    -5.4
2022 tanf	CNT	    3.095    2.930     5.6
2022 vet	CNT	    4.848    4.608     5.2
2023 mcaid	CNT	   32.598   34.334    -5.1
2023 snap	CNT	   25.572   27.022    -5.4
2023 tanf	CNT	    3.131    2.930     6.9
2023 housing	CNT	    4.072    4.342    -6.2
2024 tanf	CNT	    3.170    2.930     8.2
2025 wic	CNT	    4.884    4.620     5.7
2025 vet	CNT	    4.880    4.628     5.4
2025 housing	CNT	    4.013    4.277    -6.2
2027 snap	CNT	   25.233   27.022    -6.6
2027 tanf	CNT	    3.082    2.930     5.2
--------------------------------------------------
```
The largest count deviation is for the SNAP program in 2015 where the target recipient count is 27.968 million filing units and the actual count is 25.292 million filing units, which is about **9.6 percent below the target**.  Another example is 2024 TANF for which the actual count is about **8.2 percent above the target**.

Perhaps there is a problem with the new test, which is in the `test_benefits.py` file.  If the test is sensible, then these test results suggest that there are problems with the dual-benefit-targets algorithm implemented in the two Python scripts in  the`cps_stage4` directory.  I say that because the `extrapolation.py` script has as an explicit objective producing count deviations of no more the five percent.

### DIFF
--- a/tests/test_benefits.py
+++ b/tests/test_benefits.py
@@ -87,8 +87,8 @@ def test_extrapolated_benefits(kind, cps_benefits, puf_benefits,
     Compare actual and target extrapolated benefit amounts and counts.
     (Note that there are no puf_benefits data.)
     """
-    rtol_amt = 0.13
-    rtol_cnt = 0.15
+    rtol_amt = 0.002
+    rtol_cnt = 0.10
     dump_res = False
     # specify several DataFrames and related parameters
     if kind == 'cps':


martinholmer:
@andersonfrailey and @hdoupe and @Amy-Xu, pull request #248 is ready for your review.
Please review the new test and the extrapolation scripts to see if you can figure out why we are getting the disappointing test results presented [here](https://github.com/open-source-economics/taxdata/pull/248#issue-201037335).

@MattHJensen

hdoupe:
@martinholmer I'm trying to catch up on the work you did in PR #248. My apologies for missing the review window that you left open yesterday.

Are you saying that the extrapolated participant count is off by greater than the tolerance of 5%? If this is the case, it seems like something is going wrong between when the benefits are extrapolated and when they are saved to the results dataframe, `benefit_extrapolation`. My reasoning is that the count is within the 5% tolerance (at least the `np.allclose(...)` check says so) when I run `extrapolation.py` on the master branch.  Does that seem like a sensible assessment?

martinholmer:
@hdoupe asked in pull request #248:
> Are you saying that the extrapolated participant count is off by greater than the tolerance of 5%? 

Yes.

@hdoupe also said:
> If this is the case, it seems like something is going wrong between when the benefits are extrapolated and when they are saved to the results dataframe, benefit_extrapolation. My reasoning is that the count is within the 5% tolerance (at least the np.allclose(...) check says so) when I run extrapolation.py on the master branch. Does that seem like a sensible assessment?

As I said before, there are a couple of possible reasons for this discrepancy.  The new test could be flawed, the `extrapolation.py` script could be flawed, or the comparison being made in the script is not the same as the comparison being made in the test.  I have no idea wha't going on in the script, so as its author it seems like you are the person to resolve this.

martinholmer:
@martinholmer said in response to @hdoupe in merged pull request #248:
> As I said before, there are a couple of possible reasons for this [test] discrepancy. The new test could be flawed, the extrapolation.py script could be flawed, or the comparison being made in the script is not the same as the comparison being made in the test. I have no idea wha't going on in the script, so as its author it seems like you are the person to resolve this.

I've been talking to @andersonfrailey about the meaning of benefit-program "participants" and we're now thinking that they are individual head counts.  If that understanding is correct, then the number of participants in a filing unit would be somewhere in the [1, XTOT] range.  I guess we don't know anything more than that from the `cps_benefits.csv.gz` file, right?

Pending pull request #252 generalizes the test so that we can assume either one participant per filing unit (as in earlier versions of the test) or `XTOT` participants per filing unit (see #252 for the results using that assumption).

@hdoupe and @andersonfrailey  I'd value your thoughts on this matter.
I'm going to leave pull request #252 open until early next week.

PR #249, opened 2018-07-12
Experiment with no benefit-count targeting [DON'T MERGE]

martinholmer:
This pull request is an experiment that is not really intended for merging into the taxdata master branch.
What this branch does is generate extrapolated benefits when the benefit-count targeting logic is turned off.
The point of this experiment is to compare the extrapolated benefit counts in two different cases: 
1. when benefit counts are targeted (which is the case on the taxdata master branch) in the `cps_stage4/extrapolation.py` script, and
2. when they are not targeted in the `cps_stage4/extrapolation.py` script.

Note that the code changes in this pull request are intentionally causing `test_extrapolated_benefits` to fail.

I will post other comments in this pull request that show the results of this experiment.

### DIFF
--- a/cps_stage4/extrapolation.py
+++ b/cps_stage4/extrapolation.py
@@ -272,7 +272,7 @@ if __name__ == '__main__':
     # create extrapolated benefits
     ben = Benefits()
     for _ in range(SYEAR, LYEAR):
-        ben.increment_year(tol=0.05)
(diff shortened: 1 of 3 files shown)


martinholmer:
This pull request was never intended to be merged into the master branch, so it is being closed now.

PR #263, opened 2018-07-24
Remove arguments from runmatch.py

andersonfrailey:
This PR updates `runmatch.py` so that instead of a user specifying the path to input files the program checks for the files automatically. If the CSV version of the appropriate CPS file is found, it will read that in. If there is no CSV file but there is a .DAT version, it will create the CPS from the .DAT file. Otherwise an error is raised.

I chose to raise an OSError because, according to the Python [docs](https://docs.python.org/3/library/exceptions.html#OSError),
>This exception is raised when a system function returns a system-related error, including I/O failures such as “file not found”

cc @martinholmer

### DIFF
--- a/puf_data/StatMatch/Matching/runmatch.py
+++ b/puf_data/StatMatch/Matching/runmatch.py
@@ -7,39 +7,31 @@ from phase2 import phasetwo
 from add_cps_vars import add_cps
 from add_nonfilers import add_nonfiler
 import pandas as pd
-import argparse
+import os
 
 """
     Script to run each phase of the matching process
 """
 
 
-def match(mar_cps_path='asec2016_pubuse_v3.dat',
-          puf_path='puf2011.csv'):
-    # Add arguments for specifying path to CPS file in CSV format
-    # this will allow the program to skip the process of creating the CPS from
-    # a .DAT file.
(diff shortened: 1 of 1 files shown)


andersonfrailey:
Noticed the `FileNotFoundError` as I was closing the documentation so I've updated this PR to use that instead. I'd say it's a more accurate description that simply `IOError`.

puf_data/StatMatch/Matching/runmatch.py
```diff
@@ -7,47 +7,39 @@
...
-    # Create CPS file either from a CPS or through create_cps method
-    if args.cps is not None:
-        mar_cps = pd.read_csv(args.cps)
+def match():
+    # If there is a .CSV version of the CPS, simply read that in. Otherwise
+    # convert the .DAT file to a .CSV
+    cps_csv_path = 'cpsmar2016.csv'
+    if os.path.isfile(cps_csv_path):
+        print('Reading CPS Data')
+        mar_cps = pd.read_csv('cpsmar2016.csv')
```
martinholmer: Might want to have this line be:
```
mar_cps = pd.read_csv(cps_csv_path)
```
which is line the way you handled it below with the DAT file.

PR #252, opened 2018-07-13
Revise benefit test so that filing-unit "participants" can be XTOT or one

martinholmer:
This pull request generalizes the new extrapolated benefit test that was added in pull request #246 and modified in pull requests #248 and #250.  Prior versions of `test_extrapolated_benefits` assumed that the participant counts were filing units rather than individuals.  The problem for testing is that there is no information in the `cps_benefits.csv.gz` file about how many individuals in each filing unit are benefit-program participants.  The actual number of participants is obviously somewhere in the [1, `XTOT`] range.

The test revisions in this pull request add a `size_is_xtot` boolean switch that sets the number of benefit-program participants in each filing unit to one (when `size_is_xtot == False`) or to `XTOT` (when `size_is_xtot == True`).  When `size_is_xtot == True` and `rtol_cnt == 0.05` (rather than its normal value of 0.12), the test produces these results:
```
--------------------------------------------------
TEST RESULTS WITH rtol_amt=0.002 and rtol_cnt=0.05
                  and size_is_xtot=True
--------------------------------------------------
2015 mcare	CNT	   67.018   63.651     5.3
2015 tanf	CNT	    6.679    6.111     9.3
2016 ssi	CNT	   13.957   13.087     6.6
2016 wic	CNT	   17.639   16.547     6.6
2016 tanf	CNT	    8.073    7.479     7.9
2016 housing	CNT	    9.232    8.640     6.9
2017 ssi	CNT	   13.800   13.061     5.7
2017 snap	CNT	   52.647   49.634     6.1
2017 tanf	CNT	    8.166    7.479     9.2
2018 ssi	CNT	   14.114   13.061     8.1
2018 snap	CNT	   53.461   49.634     7.7
2018 tanf	CNT	    8.255    7.479    10.4
2018 housing	CNT	    8.998    8.460     6.4
2019 tanf	CNT	    8.351    7.479    11.7
2020 ssi	CNT	   14.009   13.192     6.2
2020 mcare	CNT	   77.593   73.778     5.2
2020 snap	CNT	   52.494   49.634     5.8
2020 tanf	CNT	    8.032    7.479     7.4
2020 housing	CNT	    8.898    8.334     6.8
2021 ssi	CNT	   14.359   13.244     8.4
2021 mcare	CNT	   79.975   75.927     5.3
2021 snap	CNT	   53.296   49.634     7.4
2021 tanf	CNT	    8.119    7.479     8.6
2022 mcare	CNT	   82.405   78.100     5.5
[...]

PR #270, opened 2018-08-03
Copy benefit growth rates to puf_stage1 directory

martinholmer:
This pull request is the first step in simplifying the benefit extrapolation logic used in the taxdata and Tax-Calculator repositories.  The new logic will construct (in taxdata) and use (in Tax-Calculator) new grow factors for each benefit, which will be included in the `growfactors.csv` file.   See [this discussion](https://github.com/open-source-economics/taxdata/pull/252#issuecomment-405990257) for details on why this simplification is being done.


martinholmer:
Merging PR #270 because there is no change in taxdata logic.  The pull request simply sets the stage for future changes in the `puf_stage1/factors_finalprep.py` script that write grow factors for each benefit to the `growfactors.csv` file.

@MattHJensen @Amy-Xu @andersonfrailey @hdoupe

PR #274, opened 2018-08-14
Add inPUF variable to puf.csv file

martinholmer:
This pull request adds a new variable to the `puf.csv` file.  The new variable is called `inPUF` and it is an integer variable with a value of either 1 (indicating that the `puf.csv` record is from the IRS/SOI PUF) or 0 (indicating that the `puf.csv` record is from the CPS).  This new variable is not used in Tax-Calculator logic, but will be useful in taxdata variable imputation work.

### DIFF
--- a/puf_data/StatMatch/Matching/runmatch.py
+++ b/puf_data/StatMatch/Matching/runmatch.py
@@ -38,10 +38,12 @@ def match():
     # Remove aggregated variables from the PUF
     puf = puf[(puf['recid'] != 999996) & (puf['recid'] != 999997) &
               (puf['recid'] != 999998) & (puf['recid'] != 999999)]
+    puf['inPUF'] = 1
 
     print('Creating CPS Tax Units')
     rets = Returns(mar_cps)
     cps = rets.computation()
+    cps['inPUF'] = 0
 
     print('CPS Tax Units Created')
     filers, nonfilers = adjfilst(cps)


andersonfrailey:
@martinholmer, I'm not sure that this new variable is necessary. The `filer` variable can be used in the same way as this would be. `filer` will be 0 only if the record is a non-filer, and those records are the only ones in PUF.csv that are 100% from the CPS. Every other record, where `filer` is equal to 1, is a result of the PUF/CPS match.

martinholmer:
@andersonfrailey said:
> I'm not sure that this new [`inPUF`] variable is necessary. The `filer` variable can be used in the same way as this would be. `filer` will be 0 only if the record is a non-filer, and those records are the only ones in `puf.csv` that are 100% from the CPS. Every other record, where filer is equal to 1, is a result of the PUF/CPS match.

I'm confused.  I thought there was puf.csv-creation logic that set `filer` equal to 1 for some CPS records.
Is that not correct?

andersonfrailey:
@martinholmer, that is correct. When creating the CPS tax units we determine which records are filers, but all CPS records determined to be filers are matched with PUF records. There are no records who have `filer == 1` and are comprised solely of data from the CPS in the final file.

martinholmer:
@andersonfrailey said:
> When creating the CPS tax units we determine which records are filers, but all CPS records determined to be filers are matched with PUF records. There are no records who have filer == 1 and are comprised solely of data from the CPS in the final file.

Thanks for the clarification.  I'll close unneeded pull request #274.

PR #284, opened 2018-09-11
Avoid Python 3.7 for the time being

martinholmer:
Some packages are not yet available for Python 3.7


hdoupe:
Agreed. I made a similar decision in PolicyBrain.

PR #287, opened 2018-09-19
Add Admin Costs

andersonfrailey:
This PR adds administrative costs to the benefits in the CPS file as specified by @Amy-Xu in [this](https://github.com/open-source-economics/C-TAM/issues/60#issuecomment-393319730) comment.

I did have a couple of questions come up while working on this though. First, since Social Security and unemployment compensation are used in the Tax-Calculator logic beyond just benefit repeal, will including these administrative costs have an adverse effect on Tax-Calculator results?

Second, in PR #273, @martinholmer added a test to ensure that medicare and medicaid benefits are the actuarial values. Now that we've added administrative costs is this test still necessary/accurate?


martinholmer:
@andersonfrailey said in PR #287:
> This PR adds administrative costs to the benefits in the CPS file as specified by @Amy-Xu in this comment.
>
> I did have a couple of questions come up while working on this though. First, since Social Security and unemployment compensation are used in the Tax-Calculator logic beyond just benefit repeal, will including these administrative costs have an adverse effect on Tax-Calculator results?

**DO NOT ADD ADMINISTRATIVE COSTS to social security and unemployment insurance because this will make the Tax-Calculator tax-liability estimates INCORRECT.**

@MattHJensen @Amy-Xu

andersonfrailey:
Removed administrative costs from social security and unemployment insurance in latest commit.

@martinholmer the test is failing for medicaid now and I just wanted to make sure that I wasn't missing some reason as to why that would be other than an actual error in the new results.

martinholmer:
@andersonfrailey,  this is the code that triggers the test error:
```
        # check that mc??? benefits are actuarial values of health insurance
        if bname == 'mcare' or bname == 'mcaid':
            ratio = float(maxben) / minpben
            expect_ratio = round(ratio)
            if not np.allclose([ratio], [expect_ratio], rtol=0, atol=0.001):
                msg = '\nCPS {}_ben ratio={:.6f} != {:.0f}'
                error_msg += msg.format(bname, ratio, expect_ratio)
```
Here is the Travis CI error message:
screen shot 2018-09-19 at 11 07 31 am

Why not change  in the allclose call`atol=0.001` to `atol=0.002`?

After all,  14.001734 is pretty close to 14.

andersonfrailey:
After upping the tolerance a bit allows all the tests to pass. This PR is ready to be merged upon review.

@Amy-Xu @martinholmer @MaxGhenis

Amy-Xu:
The function for adding administrative cost looks good to me. Then this admin cost is added to each program's benefit. Frankly I haven't thought about this step much. What is the desired form to present the admin cost? @MattHJensen 

If we add the admin costs to the benefits directly, then it's hard to separate them again. I imagine we want to have a separate column in the result tables to show the distribution of total admin cost (all programs combined). But I'm not super sure. 

I agree with Martin that the admin cost for Social Security and Unemployment Insurance should not be added to the corresponding benefit variable. But probably they could be lump summed into the total admin cost variable if that's the route we go for.

One last clarification, the administrative cost for housing currently just includes the cost for housing vouchers (HCV). We need to somehow adjust/impute to get an estimate for all three programs. There's no clear agreement in the corresponding issue yet.

MaxGhenis:
Is this PR waiting on reproducing the CPS data in Python?

andersonfrailey:
@MaxGhenis, no, this can be implemented at any point since it happens in the final prep steps.

There are a couple of points made by @Amy-Xu that I think need addressing before we can merge this PR though.

One:
>If we add the admin costs to the benefits directly, then it's hard to separate them again. I imagine we want to have a separate column in the result tables to show the distribution of total admin cost (all programs combined). But I'm not super sure.

The solution to this, IMO, depends on user needs. If it's not important to split up actual benefits from admin costs, then we can just add on the admin costs and it's no big deal.

Alternatively if they do need to be separated then we'll need to create a new admin cost variable for each benefit variable. The downside to this is that's more variables we need to add to the CPS and to Tax-Calc. That being said, I think it makes more sense to keep benefits and admin costs separate.

>One last clarification, the administrative cost for housing currently just includes the cost for housing vouchers (HCV). We need to somehow adjust/impute to get an estimate for all three programs. There's no clear agreement in the corresponding issue yet.

We could always merge this PR and come back to this, but it is something to keep in mind.

---

Both of these issues have made me question whether or not this PR should be merged. It's my understanding that the goal of this is to allow users to include administrative savings associated with a repeal of benefit programs in an analysis of a policy such as replacing current transfer programs with a UBI. This is certainly something that they should do, but is the micro-data itself the place to hold those administrative costs or should we leave it to the users to add after the micro-level analysis?

I'm inclined to favor leaving adding admin costs to the user simply because I'm unsure what the benefit of including them in the micro-data is. I worry that this could cause confusion for users who are only looking to add up total benefits and maybe look at their distribution, especially if we don't separate admin costs with benefits received. And given that benefit value is included in `expanded_income` in taxcalc, it doesn't make sense to me to include admin costs that individuals don't actually benefit from in that calculation.

With regard to including all admin costs for housing programs, if we merge this PR without solving how we include all admin costs, then we'll be underestimating those costs without having a direct way of letting users know. And it might be better to let our users determine how they want to calculate those costs, rather than doing it ourselves.

I would, of course, love to hear other's opinions on this.

cc @MattHJensen @martinholmer

PR #294, opened 2018-12-06
CBOUpdate2

lmm8ka:
updated CBO baseline values using CBO 10 year projections, updating from August 2016 projections to the August 2018 projections. Aug 2018 includes actual values for the period between Aug 2016-2018. Also includes Data from IRS Publication 6961 from Fall 2018, replacing data from Publication 6187, Fall 2016. Also updating OASI Trust Fund Annual Trustees Report Operations of the OASI Trust Fund, Fiscal Years 2011-15 to be replaced with OASI Trust Fund Annual Trustees Report FY 2013-2027. And lastly CBO Unemployment Compensation Projections from March 2016 iteration to include the April 2018 data. All of this updated the CBO baseline values from the updated 2016-2018 real values into the 2018-2028 projected values.

PR #295, opened 2019-01-04
Complete CBO and SOI Updates

andersonfrailey:
This PR fixes this issues I referenced in PR #294. I needed to add on the new SOI estimates, fixed the values for CPIM in the CBO baseline, and then re-ran everything. For the most part, there was not a huge change in tax liability projections except for 2018, as can be seen in [this notebook](https://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/Add2028Projections.ipynb). I believe that the drop in tax liabilities for 2018 can be traced back to lower projections for dividend and wage income.


martinholmer:
@andersonfrailey,  You didn't mention in your initial comment on PR #295 that the proposed changes cause two serious test failures:

1.  The minimum PUF INT ratio for 2016 is -0.0065.  Not only is this below the test's plausible minimum of +0.2, but it is negative.  It would seem a negative 2016 ratio is going to imply negative interest amounts in Tax-Calculator for all years beginning in 2016.  This is a very serious problem.  Why is the PUF INT ratio negative for one or more AGI groups in 2016?

2. The minimum filing-unit weight in 2016 is a very large (in absolute value) negative number:  -104886940419296.  A negative weight makes no sense.  I thought the linear program that calculates weights restricts them to be non-negative.  Is that not correct?  The existence of such a large negative weight in the proposed changes suggests a very serious problem.  How can the PUF data preparation procedure generate negative weights?

My biggest worry about PR #295 is that it even exists as a GitHub pull request.   A pull request should never be opened until all the tests pass.   Did all the tests pass on the local computer but fail on GitHub?  Or did the local tests not get run?

I don't care what the notebook results are when there are glaring test failures and absurd calculated parameters such as negative ratios and negative weights.

@MattHJensen @lmm8ka

andersonfrailey:
@martinholmer this is my mistake. I got a little antsy/distracted while getting this PR ready and didn't give it a thorough enough review before opening, including forgetting to run the tests and check back in on it after the tests ran. I'm going to close this, fix the mistakes, and open a fresh PR.

PR #298, opened 2019-01-15
Usability Upgrades

andersonfrailey:
This PR makes a number of changes aimed at making the TaxData project easier to use and update. None of them change the files that are produced by TaxData

First, a number of code comments that are no longer relevant were deleted.

Second, I used the `os` package to handle all of the file paths. The goal here is to avoid any issues with file reading if a user, for whatever reason, tries to run individual files rather than use the `make *` commands. The paths weren't a major problem, but there have been times in the past where users reported errors to me that ultimately were the result of them doing something like `python puf_stage1/stage1.py` from the top of the TaxData directory. Ideally users will always just run the `make *` commands like the documentation instructs, but just in case they don't these changes will prevent any major errors.

Third, I replaced references to specific years in the `stage1` files to instead reference start and end year variables that are defined at the top of individual files. Now when we update the CBO baseline and project future into the future, this is the only variable that needs to be updated.

Finally, I've replaced the outdated .TXT documentation files for updating the CBO baseline and SOI estimates with updated markdown files. They should be much more clear and detailed, making it easier to keep the repo up to date.


andersonfrailey:
Given that there are no substantive changes in this PR, I'll merge it at the end of the day if there are no objections.

PR #308, opened 2019-03-04
Partially Automate CBO Baseline Updates

andersonfrailey:
This PR adds a new script `puf_stage1/updatecbo.py` that automates parts of the `CBO_baseline.csv` updating process. It can pull all of the needed data from the CBO 10-year economic projections file, the CBO revenue projections, and uses the BLS API to update CPI-M. I say it's a partial automation because three variables cannot be easily automated: `RETS`, `SOCSEC`, and `UCOMP` all either come from a PDF or a website that cannot be reliably scraped. The documentation has been updated to reflect this and the new script will print out a message after it has finished reminding the user to update those manually. Additionally, the file needs to be updated to point to the new CBO files each time there's a new release.

Though there is still a non-trivial amount of work that needs to be done to update `CBO_baseline.csv`, this file handles a large chunk of tasks associated with the updates and significantly reduces the likelihood of user error in the updating process.

I did not add this script to the make file process because of all the manual work that needs to accompany its use.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -11,3 +11,5 @@ dependencies:
 - pytest
 - pulp
 - tqdm
+- requests
+- lxml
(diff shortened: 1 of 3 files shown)


andersonfrailey:
Given that there has been no objection to adding these scripts, I'll be merging this PR at the end of the day.

PR #314, opened 2019-05-24
Create CPS file with Python Scripts

andersonfrailey:
I want to start by emphasizing that this is still very far from completion. Looking at the weighted totals it's obvious that there is still work to be done. As I mentioned in the Tax-Calculator maintainers call last night, I wanted to open this PR so that others can see what I'm doing and offer feedback/suggestions as I go. With all that being said, here's a bit on what's going on.

To start, I download the CPS files from [NBER](https://www.nber.org/data/current-population-survey-data.html). I wrote a few scripts to handle reading the .DAT files and converting them to .CSV files: `transform_sas.py`, `template.txt`, `cpsmar*.py`. I'll be happy to explain what they each do if there's interest. I'm using the March 2013, 2014, and 2015 CPS supplements to be consistent with our current data.

After that, I take each of the CPS files and form tax units. I only look within households for tax unit formation.

Starting with the reference person for each family, I loop through everyone in the household. If they are not already part of a tax unit (as a spouse, dependent, or head of the unit) I form a new tax unit with them. I then search through everyone in the household to see if they have a spouse or dependents to claim. I also check all dependents to see if they are required to file. After all this is done, the tax units are exported.

I take a DataFrame with all of the units and run it through a final prep step to clean up the file, and the result is a CSV that should be usable with taxcalc (haven't tried it yet).

I've written a short script `validate.py` that does some very basic testing of the file and the results are promising. There are two things it's checking for: 1) has everyone in the household been accounted for and 2) is income in the household being counted correctly. 1) is checked by making sure the number of people appearing in the `nu18`, `n1820`, and `n21` variables are equal to the actual number of people in the household within those age ranges. 2) is checked by looking at a number of income variables <sup>1</sup> and asserting that the total in every tax unit in that household is less than or equal to the total amount in the household. I allow it to be less than the total in the house because there could be some dependents who have a little income, but not enough to have to file. All of these checks pass with no errors.

In terms of weighted totals, there are mixed results. Some variables, like wages, are within 2% of the PUF. Others, like farm income, are significantly off. Given that the basic validation testing I've done has passed, It's possible that these issues could be solved by reweighing. I will try and figure out the best way for sharing the current weighted totals and comparisons to the PUF and CPS files.

Next steps are to try and replicate the imputations currently being done for the CPS so that we have all the same variables. Add the benefit data from C-TAM. And do general code improvements.

<sup>1</sup> Wages, interest income, self-employment income, farm income, dividend income, unemployment compensation, pensions, and alimony

cc @MattHJensen @MaxGhenis @martinholmer


MaxGhenis:
>2) is checked by looking at a number of income variables 1 and asserting that the total in every tax unit in that household is less than or equal to the total amount in the household. I allow it to be less than the total in the house because there could be some dependents who have a little income, but not enough to have to file.

Do the scripts create tax units for nonfilers, e.g. with a `filer` flag set to `False`?

andersonfrailey:
@MaxGhenis at this point they do not. My reservation with adding a `filer` flag is that as filing requirements change it can be misleading. I think a `filer` variable would be more useful if it was determined in tax-calc where it could be determined based on the year and policy the model is being run for.

MaxGhenis:
OK why discard dependents who don't have enough to file? Should that instead be done in taxcalc (if at all)? I thought one of the valuable aspects of the CPS file is capturing all income.

andersonfrailey:
Are you proposing making all dependents into their own tax unit, or just those with income?

MaxGhenis:
How does the SAS-produced CPS file do it?

martinholmer:
@andersonfrailey,  What you're doing in taxdata #314 is a good idea.  Thank you for undertaking this project.

My main questions is about this statement:
>  I'm using the March 2013, 2014, and 2015 CPS supplements to be consistent with our current data.

Are you doing this only to be able to compare results from the current `cps.csv.gz` file or do you see this as the end product of your work?  Is makes sense to me to check the new Python approach with the old SAS approach by having the two approaches produce the same data file.  But once you have confidence in the logic of your Python approach, wouldn't it make sense to switch away from the Mar13+Mar14+Mar15 data?  Wouldn't more recent data be better?  And why exactly do we need three Mar files?   Why not just one (the most recent CPS available)?  I never understood why John made such a large file; it doesn't add anything to Tax-Calculator as far as I can see, and it does impose costs in terms of larger memory requirements and slower execution.

andersonfrailey:
@MaxGhenis 
>How does the SAS-produced CPS file do it?

The SAS-produced file only includes dependents as tax units if they are required to file themselves. All others are counted up and represented by the tax unit claiming them, but any potential income they may have is not included anywhere.

@martinholmer 
>Are you doing this only to be able to compare results from the current cps.csv.gz file or do you see this as the end product of your work?

I'm doing this just to compare my results to the current `cps.csv.gz` file.

> Is makes sense to me to check the new Python approach with the old SAS approach by having the two approaches produce the same data file. But once you have confidence in the logic of your Python approach, wouldn't it make sense to switch away from the Mar13+Mar14+Mar15 data? Wouldn't more recent data be better?

Yes. I believe that it would be better to use a more recent file once I have the Python approach ironed out. The CPS we end up picking should be a discussion between the TaxData, Tax-Calculator, and C-TAM projects (cc @Amy-Xu ).

>And why exactly do we need three Mar files? Why not just one (the most recent CPS available)? I never understood why John made such a large file; it doesn't add anything to Tax-Calculator as far as I can see, and it does impose costs in terms of larger memory requirements and slower execution.

My understanding is that three files were used in an effort to increase the sample size. I don't believe that it is necessary to continue using three separate files. I'll compare what we get when using one file versus three so we can get a firm view of what, if any, advantage there is to using three files.

MaxGhenis:
>The SAS-produced file only includes dependents as tax units if they are required to file themselves. All others are counted up and represented by the tax unit claiming them, but any potential income they may have is not included anywhere.

OK thanks. It'd be useful to capture all income at some point to evaluate reforms that affect filing rules, but makes sense to keep that out of this PR.

>My understanding is that three files were used in an effort to increase the sample size.

FWIW there's precedent for this in the [Supplemental Poverty Measure](https://www.census.gov/content/dam/Census/library/publications/2018/demo/p60-265.pdf), where state-level poverty estimates are only published using 3-year estimates, though most of the SPM and other similar analyses I've seen just use one year. Whether this is analytically better for taxcalc seems like an empirical question, but IMO a one-year file would be easiest, also because it's simpler to merge with the original CPS.
image

andersonfrailey:
My latest commits build on the work that's been done up to this point. In them I've added the C-TAM benefits, enhanced the validation tests, and added state level income targeting*. At this point I'm still working with the 2013, 2014, and 2015 CPS files so that we can compare to the current version of `cps.csv.gz`.

When comparing totals between the Python generated CPS, the PUF, and the current CPS file, there are still some odd results. You can download a comma-separated text file to see this
[here](https://github.com/PSLmodels/taxdata/files/3339835/pycpstotals.txt). EIC, elderly_dependents, and the benefit variables particularly stand out.

It would be helpful for me to have the logic I'm using to determine who is a dependent and who is EIC eligible reviewed. The code I wrote for both could definitely be improved.

The next step is to add the imputations that are on `cps.csv.gz`, but not in the Python generated CPS yet.

Despite some odd aggregate numbers, I'm fairly confident in what I have so far. In the validation tests, I make sure that everyone in the household is accounted for in one of the age variables (`nu18`, `n1820`, and `n21`) and that all income and benefits accrued by the household gets counted. Each of these tests pass without issue.

*For state level targeting, I take [state level IRS data](https://www.irs.gov/statistics/soi-tax-stats-historic-table-2) on the amount of certain income variables and adjust incomes in the CPS so that these totals are hit for each state. We do this when making the current version of `cps.csv.gz`. I haven't yet restricted this targeting to be done using just those who would file taxes.

cc @martinholmer @MaxGhenis @MattHJensen

cps_data/pycps/finalprep.py
```diff
@@ -0,0 +1,116 @@
...
+    # recode blind variables
+    data["blind_head"] = np.where(
+        data["blind_head"] == 1, 1, 0
+    )
+    data["blind_spouse"] = np.where(
+        data["blind_spouse"] == 1, 1, 0
+    )
+
+    # rename variables
+    renames = {
```
MaxGhenis:
```suggestion
    RENAMES = {
```

cps_data/pycps/pycps.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+def eic_eligible(person: dict) -> int:
+    """
+    Function to determine if a dependent is an EIC eligible child
+    df: DataFrame of just dependents
+    https://www.irs.gov/credits-deductions/individuals/earned-income-tax-credit/qualifying-child-rules
+    """
+    # relationship test
+    relationship = ((person["a_exprrp"] == 5) |
```
MaxGhenis:
```suggestion
    relationship = person.a_exprrp.isin([5, 7, 9, 11])
```

cps_data/pycps/pycps.py
```diff
@@ -0,0 +1,277 @@
...
+                    (person["a_exprrp"] == 7) |
+                    (person["a_exprrp"] == 9) |
+                    (person["a_exprrp"] == 11))
+    # age test
+    eic_max_age = 19
+    if person["a_ftpt"] == 1:
+        eic_max_age = 24
+    # person is between 1 and the max age
+    age = 0 <= person["a_age"] <= eic_max_age
```
MaxGhenis:
```suggestion
    age = (0 <= person["a_age"] <= eic_max_age)
```

MaxGhenis: (clarity only)

cps_data/pycps/pycps.py
```diff
@@ -0,0 +1,277 @@
...
+    cps["alimony"] = alimony
+    cps["pensions_annuities"] = pensions_annuities
+
+    # flags used when creating a unit
+    cps["p_flag"] = False  # primary flag
+    cps["s_flag"] = False  # spouse flag
+    cps["d_flag"] = False  # dependent flag
+    cps["hhid"] = cps["h_seq"]
+    # calculate earned and unearned income
+    earned_inc_vars = [
```
MaxGhenis:
```suggestion
    EARNED_INC_VARS = [
```

MaxGhenis: also below

cps_data/pycps/targeting.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    Read state level income information and adjust CPS data accordingly
+    """
+    state_data = pd.read_csv(state_data_link, index_col="STATE",
+                             thousands=",")
+    # only use aggregate data
+    state_data = state_data[state_data["AGI_STUB"] == 0].copy()
+
+    # map fips codes and to state abbreviations
+    fips_dict = {'AK': 2, 'AL': 1, 'AR': 5, 'AZ': 4, 'CA': 6, 'CO': 8, 'CT': 9,
```
MaxGhenis:
```suggestion
    FIPS_DICT = {'AK': 2, 'AL': 1, 'AR': 5, 'AZ': 4, 'CA': 6, 'CO': 8, 'CT': 9,
```

cps_data/pycps/targeting.py
```diff
@@ -0,0 +1,66 @@
...
+                 'ID': 16, 'IL': 17, 'IN': 18, 'KS': 20, 'KY': 21, 'LA': 22,
+                 'MA': 25, 'MD': 24, 'ME': 23, 'MI': 26, 'MN': 27, 'MO': 29,
+                 'MS': 28, 'MT': 30, 'NC': 37, 'ND': 38, 'NE': 31, 'NH': 33,
+                 'NJ': 34, 'NM': 35, 'NV': 32, 'NY': 36, 'OH': 39, 'OK': 40,
+                 'OR': 41, 'PA': 42, 'RI': 44, 'SC': 45, 'SD': 46, 'TN': 47,
+                 'TX': 48, 'UT': 49, 'VA': 51, 'VT': 50, 'WA': 53, 'WI': 55,
+                 'WV': 54, 'WY': 56}
+    # map income variables in the CPS and IRS data
+    # TODO: Add imputed variables
+    var_map = {
```
MaxGhenis:
```suggestion
    VAR_MAP = {
```

cps_data/pycps/targeting.py
```diff
@@ -0,0 +1,66 @@
...
+    factor_dict = {}
+
+    # loop through each state and variable
+    for var, cps_vars in var_map.items():
+        factor_dict[var] = []
+        print(var)
+        for state, fips in fips_dict.items():
+            sub_cps = cps[cps["fips"] == fips]
+            target = state_data[var][state] * 1000  # scale up IRS data
+            cps_uw_total = sub_cps[cps_vars].sum(axis=1)
```
MaxGhenis: Add a comment noting this is a vector with the total for each tax unit (I was confused at first).

cps_data/pycps/taxunit.py
```diff
@@ -0,0 +1,195 @@
+INCOME_TUPLES = [
```
MaxGhenis: Philosophical question about this file: Why make a new `TaxUnit` object, rather than a DataFrame?

andersonfrailey:
Thanks for the comments, @MaxGhenis. I'll be working through them in upcoming commits

cps_data/pycps/benefits.py
```diff
@@ -0,0 +1,119 @@
...
+    ss = pd.read_csv(ss_path, usecols=["ss_val", "peridnum"])
+    ss = ss.rename(columns={"ss_val": "ss_impute"})
+
+    housing_path = Path(data_path, f"Housing_Imputation_logreg_{year}.csv")
+    housing = pd.read_csv(housing_path,
+                          usecols=["fh_seq", "ffpos", "housing_impute"])
+
+    tanf_path = Path(data_path, f"TANF_Imputation_{year}.csv")
+    tanf = pd.read_csv(tanf_path, usecols=["peridnum", "tanf_impute"])
+    # drop duplicated people in tanf
```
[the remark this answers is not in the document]
andersonfrailey: @Amy-Xu could provide a better answer to this than I could

cps_data/pycps/taxunit.py
andersonfrailey: This was mostly to make it easier for me to conceptualize what was happening. Treating the tax unit as its own object made it easier for me to form the logic of creating a tax unit, without severely complicating transforming that `TaxUnit` object into a row in the final DataFrame.

cps_data/pycps/benefits.py
```diff
@@ -0,0 +1,118 @@
...
+    snap = read_ben("SNAP_Imputation", ["h_seq", "snap_impute"])
+    ssi = read_ben("SSI_Imputation", ["ssi_impute", "peridnum"])
+    ss = read_ben(
+        "SS_augmentation", ["ss_val", "peridnum"]
+    ).rename(columns={"ss_val": "ss_impute"})
+    housing = read_ben("Housing_Imputation_logreg_",
+                       ["fh_seq", "ffpos", "housing_impute"])
+    tanf = read_ben("TANF_Imputation_", ["peridnum", "tanf_impute"])
+    # drop duplicated people in tanf
+    tanf = tanf.drop_duplicates("peridnum")
```
MaxGhenis:
```suggestion
    tanf.drop_duplicates("peridnum", inplace=True)
```

cps_data/pycps/benefits.py
```diff
@@ -0,0 +1,118 @@
...
+    # ... remove TANF and WIC from other_ben total
+    tanf_total = (data["tanf_ben"] * data["s006"]).sum()
+    wic_total = (data["wic_ben"] * data["s006"]).sum()
+    other_ben_total = other_ben["2014_cost"].sum() - tanf_total - wic_total
+    # ... divide by the weight to account for weighting in Tax-Calculator
+    data["other_ben"] = (data["ratio"] * other_ben_total / data["s006"])
+
+    data["housing_ben"] *= 12
+    # Convert benefit data to integers
+    ben_vars = [
```
MaxGhenis:
```suggestion
    BEN_VARS = [
```

cps_data/pycps/benefits.py
```diff
@@ -0,0 +1,118 @@
...
+        "mcaid_ben", "mcare_ben", "ssi_ben", "snap_ben", "vet_ben", "tanf_ben",
+        "wic_ben", "housing_ben", "e02400", "e02300", "other_ben"
+    ]
+    for var in ben_vars:
```
MaxGhenis:
```suggestion
    for var in BEN_VARS:
```

cps_data/pycps/finalprep.py
```diff
@@ -0,0 +1,116 @@
...
+        'e24515', 'e24518', 'e26270', 'e27200', 'e32800', 'e58990', 'e62900',
+        'e87530', 'elderly_dependents', 'f2441', 'f6251', 'n24',
+        'nu05', 'nu13', 'nu18', 'n1820', 'n21', 'p08000', 'p22250', 'p23250',
+        'p25470', 'p87521', 's006', 'e03210', 'ssi_ben', 'snap_ben',
+        'vet_ben', 'mcare_ben', 'mcaid_ben', 'oasdi_ben', 'other_ben',
+        'h_seq', 'ffpos', 'fips', 'a_lineno', 'tanf_ben', 'wic_ben',
+        'housing_ben', "linenos"
+    ]
+
+    drop_vars = []
```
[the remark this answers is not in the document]
MaxGhenis: You can now delete lines 31-34

cps_data/pycps/pycps.py
```diff
@@ -0,0 +1,274 @@
...
+    UNEARNED_INC_VARS = [
+        "int_val", "div_val", "rtm_val",
+        "alimony", "uc_val"
+    ]
+    cps["earned_inc"] = cps[EARNED_INC_VARS].sum(axis=1)
+    cps["unearned_inc"] = cps[UNEARNED_INC_VARS].sum(axis=1)
+    # add apply create_unit with a progress bar
+    tqdm.pandas(desc=str(year))
+    units = cps.groupby("h_seq").progress_apply(create_units, year=year - 1)
+    TAX_UNITS = []
```
MaxGhenis: I don't think this is a constant as the capitalization suggests

andersonfrailey:
BTW, I'm planning on using the same weighting methodology as we currently use for the CPS file. If anyone else has another method they'd like to try, I'd be happy to experiment with it.

MaxGhenis:
>BTW, I'm planning on using the same weighting methodology as we currently use for the CPS file. If anyone else has another method they'd like to try, I'd be happy to experiment with it.

Just throwing out https://github.com/PSLmodels/taxdata/issues/323 to keep the weights for tax units in the same household equal, so that they can be reaggregated to households.

andersonfrailey:
I just pushed a bunch of new code related to imputations. I've just copied our current imputation methods, but I believe they could be improved on. This is something that should be explored in the future.

andersonfrailey:
Latest commit just refactors the logic to create the tax units by doing as much as possible with dictionaries, instead of DataFrames. This has brought the runtime of the tax unit creation process down to around 20 seconds per year. All other logic is unchanged.

andersonfrailey:
I was able to use the linear programming model @yimengyin16 used in his presentation linked to in issue #326 and generate weights for the python generated CPS file. the new files through taxcalc to see how everything compares to the current PUF and CPS file. Will do later today. 

I'm mostly confident in the logic that creates the tax unit, but some of the new benefit totals seem odd to me.  I've attached output from the test failure and you can see which ones look most off. I'm most concerned with medicare, medicaid, SNAP, TANF, and housing. Later today I'm going to dig more into the C-TAM data to see if things like the maximum value for medicare are legit or a bug with the new code. 
Screen Shot 2019-12-29 at 11 19 12 AM

Overall though, I'd say things are looking up for this project. As always, open to any feedback y'all have.

andersonfrailey:
Here's a [notebook](https://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/comparison.ipynb) that does a short comparison of the python generated CPS file and the current CPS file and PUF. The most obvious difference is that the python generated CPS shows consistently lower income tax liabilities than the other two files.

I also looked into the high benefit numbers and they match up with C-TAM data. And the final CPS file also passes all of the validation tests I wrote for checking income and benefit totals in each tax unit.

As always, happy to hear any feedback/suggestions.

MaxGhenis:
Thanks @andersonfrailey, what do you think is driving the lower income tax in the new CPS? For 2019, it looks like $1.1T vs. $1.4T in the old CPS and $1.7T in the PUF. Payroll tax totals less than the old CPS but similar to the PUF so that doesn't seem like an issue.

Cell 19 shows that the new CPS tax units skew lower-income than the new ones, but I couldn't tell how much of the income tax difference that accounts for.

Could you share the grand total weight for each of the 3 sources? Cells 19-20 suggest they might differ by 10% or more.

If they do differ substantially, it could be interesting to see how a different version of the CPS compares, limiting to filers and scaling up the weights to match the PUF total weight.

andersonfrailey:
@MaxGhenis Here are the total numbers:

|         | Combined     | IITax        | PTax         |
|---------|--------------|--------------|--------------|
| PUF     | 39873.002638 | 23445.090884 | 16427.911754 |
| Old CPS | 37745.819397 | 20543.109070 | 17202.710328 |
| New CPS | 32697.887604 | 16171.936765 | 16525.950839 |

So the ten year aggregate income tax totals are very far off. 

What I've noticed is that the biggest difference in income tax liability between the new and old (current) CPS file is in the $200-500K expanded income bin. That gap makes up about 47% of the total gap for 2019.
visualization

But when you look at total taxable income, the new CPS file actually lines up pretty well with both the PUF and the old CPS.

visualization-2

I realized last night that there's an issue with how I did state level targeting. There's one part of the process where we use state level IRS data to scale up or down people's income to meet the totals reported by the IRS in each state. This is something we do with the current CPS as well. But, rather than only using the people who we've determined to be filers to determine these scaling factors, I use everyone in the CPS. It's possible that this is causing the scaling factors to be too small.

I have no idea if this will actually solve the problem, but I'll re-run everything after the switch has been made and we'll see what happens.

andersonfrailey:
Update: It did not change total taxes very much so there's still work to do. But I am excited that all the files needed to create the CPS file and weights run in ~15 minutes rather than 6+ hours!

MaxGhenis:
Could you share the total number of tax units in the 3 files?

andersonfrailey:
Screen Shot 2020-03-16 at 10 29 14 PM
Here's info on the number of tax units, @MaxGhenis.

There are a ton more filers in the python generated version of the CPS, and I'm still trying to figure out what exactly is causing that. I think I'm much more strict about how tax units are formed than the SAS code has been. For example, there's a  household in the 2015 CPS containing the reference person, their two kids, and an unmarried partner. In the SAS produced CPS, this is one tax unit with the unmarried partner filing as the head of household because they're the only one with income. In the Python CPS, this is two units, both single filers: the first is the reference person and their children (there's no income in this unit), and the unmarried partner. I've been doing some research on the exact rules for forming a tax unit and how I need to change my code accordingly.

In the meantime, I fixed an issue that marked married people whose spouse absent from the file as single filers, rather than joint (according to the documentation, being absent from the file is distinct from being separated). I've added the income distribution adjustments that we've been doing, and added state taxes after credits to the file as `e18400`.

These fixes closed some of the gap between combined tax liabilities, but we're still not looking great:

| | 2019             | 2020    | 2021    | 2022    | 2023    | 2024    | 2025    | 2026    | 2027    | 2028    | 2029    |
|------------------|---------|---------|---------|---------|---------|---------|---------|---------|---------|---------|---------|
| puf_combined     | 2863.51 | 2988.47 | 3105.39 | 3221.69 | 3351.85 | 3500.75 | 3656.24 | 4018.88 | 4198.20 | 4387.00 | 4581.02 |
| old_cps_combined | 2615.40 | 2774.24 | 2917.89 | 3052.38 | 3191.31 | 3338.06 | 3482.61 | 3832.32 | 4001.00 | 4178.88 | 4361.73 |
| new_cps_combined | 2433.51 | 2550.49 | 2660.41 | 2778.79 | 2895.14 | 3019.66 | 3152.28 | 3436.78 | 3585.61 | 3741.84 | 3904.89 |

andersonfrailey:
I think I've got it. Played with the imputations and the state targeting a bunch and now the 10 year total tax liability is very close to the current CPS. Total income tax liability is about 0.5% less than the current CPS and payroll tax liability is about 4% less than the current CPS, but right on payroll tax liability in the PUF. Here's a plot of them all
combined

Not sure why the build is failing in Travis, but all the tests pass locally.

cc @MattHJensen @MaxGhenis @Peter-Metz

MaxGhenis:
Awesome! The Travis [build](https://travis-ci.com/github/PSLmodels/taxdata/builds/171191092) is failing on the environment:
```
$ conda env update -f environment.yml
SpecNotFound: Can't process without a name
The command "conda env update -f environment.yml" failed and exited with 1 during .
```
I think this is because the `pip` portion needs to have the dash and indents ([docs](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#create-env-file-manually)); a correct specification (also avoiding the warning requesting `pip` to be installed) would replace:

https://github.com/PSLmodels/taxdata/blob/afd2621ae6b930258f08056967515f1cb4d00e68/environment.yml#L20-L22

with
```
- pip                                                                                                                                                                                                                                        
- pip:                                                                                                                                                                                                                                       
  - rpy2                                                                                                                                                                                                                                     
  - tzlocal 
```
However, this throws this error on my machine, and might do so on Travis too:
```
Pip subprocess error:
    ERROR: Command errored out with exit status 1:
     command: /home/mghenis/anaconda3/envs/taxdata-dev/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-7ndz7san/rpy2/setup.py'"'"'; __file__='"'"'/tmp/pip-install-7ndz7san/rpy2/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-oqfbuzdl
         cwd: /tmp/pip-install-7ndz7san/rpy2/
    Complete output (2 lines):
    cffi mode: CFFI_MODE.ANY
    Error: rpy2 in API mode cannot be built without R in the PATH or R_HOME defined. Correct this or force ABI mode-only by defining the environment variable RPY2_CFFI_MODE=ABI
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```
Can you use the `conda` installers for `rpy2` and `tzlocal` instead? That worked for me, after adding `r` to the list of channels (`rpy2` isn't on `conda-forge`).

MaxGhenis:
You might still run into problems in the build though, since the `rpy2` section imports the `censReg` R package, which seems only to be on CRAN (probably won't be installed in Travis). `censReg` is used for the tobit regressions for [imputing charitable deductions](https://github.com/PSLmodels/taxdata/pull/314/files?file-filters%5B%5D=.py#diff-5c433e930b1216b744733b6bc2d5179a) without producing negative valued imputations (the only use of `rpy2`). Alternatives could be using Python code for tobit regressions (e.g. [this](https://github.com/jamesdj/tobit) is based on `censReg` but isn't a full package), or just using OLS and truncating at zero post-hoc. Either way we'd probably want to explore alternative imputation approaches in the future.

andersonfrailey:
Thanks for the tips, @MaxGhenis! I think I'm going to remove those two packages for now. I ended up using the parameters John used for imputations rather than the ones from censreg. I also had a bunch of trouble installing rpy2 locally recently and figured we could stick with John's until that got figured out.

andersonfrailey:
I feel comfortable merging this. Happy to answer any lingering questions about it before doing so.

cps_data/pycps/helpers.py
```diff
@@ -0,0 +1,108 @@
...
+X_VARS = ["lntot_inc", "joint_filer", "fam_size", "agede", "constant"]
+
+
+class FilingParams(Parameters):
+    defaults = Path(CUR_PATH, "filing_rules.json").open("r").read()
+    label_to_extend = "year"
+    array_first = True
+
+
+FILINGPARAMS = FilingParams()
```
MaxGhenis, line 24: not a constant for uppercase

cps_data/pycps/helpers.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    Read in all C-TAM imputed benefits. Convert them to dictionaries for
+    faster lookups
+    """
+    def read_ben(path_prefix, usecols, index_col=None):
+        path = Path(DATA_PATH, path_prefix + str(year) + ".csv")
+        return pd.read_csv(path, usecols=usecols, index_col=index_col)
+
+    # Set global variables so they can be accested later
+    global MCAID, MCARE, VB, SNAP, SSI, SS, HOUSING, TANF, UI, WIC
```
MaxGhenis, line 47: these should be lowercase

cps_data/pycps/helpers.py
```diff
@@ -0,0 +1,108 @@
...
+
+class FilingParams(Parameters):
+    defaults = Path(CUR_PATH, "filing_rules.json").open("r").read()
+    label_to_extend = "year"
+    array_first = True
+
+
+FILINGPARAMS = FilingParams()
+# this index is used to access the filing parameter values for the correct year
+PUF_YR_IDX = PUF_YEAR - FILINGPARAMS.label_grid["year"][0]
```
MaxGhenis, line 26: this and below should be lowercase

cps_data/pycps/impute.py
```diff
@@ -0,0 +1,262 @@
...
+    # find log of specified variables
+    for var in LOG_VARS:
+        # use lowercase because CPS uses lowercase at this stage
+        data[f"ln{var.lower()}"] = log(data, var.lower())
+
+    data["joint_filer"] = np.where(
+        data["mars"] == 2, 1, 0
+    )
+    # cap family size at 5 to match with the PUF
+    # data["fam_size"] = np.minimum(data["XTOT"], 5)
```
MaxGhenis, line 88: is family size being capped elsewhere? if not could you uncomment the code, and possibly leave a comment re: why it's not being done?

cps_data/pycps/impute.py
```diff
@@ -0,0 +1,262 @@
...
+
+    # add imputed capital gains and IRA distributions to total income
+    data["tot_inc"] += data["CGAGIX"] + data["TIRAD"]
+    data["lntot_inc"] = log(data, "tot_inc")
+
+    # DPAD imputations
+    dpad_indicator = np.where(
+        (data["e00900"] > 0) | (data["rents"] > 0), 1, 0
+    )
+    dpad_bins = [
```
MaxGhenis, line 194: uppercase this and other constants in this file (dpad_probs, dpad_bases)

cps_data/pycps/impute.py
```diff
@@ -0,0 +1,262 @@
...
+    for var, params in IMPUTATION_PARAMS.items():
+        data[var] = impute(data, **params)
+
+    # tobit models
+    TOBIT_XVARS = [
+        "lntot_inc", "joint_filer", "fam_size", "agede", "constant"
+    ]
+
+    data["CHARITABLE"] = tobit(
+        data, ols_betas["char_ols"], TOBIT_XVARS, 48765.45, 1.
```
MaxGhenis, line 180: Where this 48765.45 from?

And in general there are a lot of hard-coded numbers in this file. Are they documented? Ideally I think these would be their own defined constants rather than entered in functions, or even better would be loading them from CSVs.

cps_data/pycps/splitincome.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def split_income(data):
+    """
+    Split up income variables
+    """
+    np.random.seed(79)
+
+    # Qualified dividends
+    all_qualified_prob = 0.429  # % of units where all dividends are qualified
```
MaxGhenis, line 14: uppercase (others too in this file)

cps_data/pycps/targeting.py
```diff
@@ -0,0 +1,85 @@
...
+    # map income variables in the CPS and IRS data
+    # TODO: Add imputed variables
+    VAR_MAP = {
+        "A00200": ["e00200p", "e00200s"],
+        "A00300": ["e00300"],
+        "A00600": ["divs"],
+        "A00650": ["e00650"],
+        "A00900": ["e00900p", "e00900s"],
+        "A02300": ["e02300"],
+        # "A01000": ["CGAGIX"],
```
MaxGhenis, line 36: why is this commented?

cps_data/pycps/validation.py
```diff
@@ -0,0 +1,107 @@
...
+    if elderly_deps_data > elderly_deps_cps:
+        record_error(
+            "elderly_dependents", h_seq, elderly_deps_data,
+            elderly_deps_cps, year
+        )
+        num_errors += 1
+
+    # compare income variables
+    # add up income totals in the raw cps
+    # cps_totals = dict(
```
MaxGhenis, line 77: why commented?

cps_data/pycps/validation.py
```diff
@@ -0,0 +1,107 @@
...
+        cps_sum = sum([p[_cps] for p in cps])
+        tc_sum = data[_tc].sum()
+        # it's acceptable for the sum in the CPS to be larger
+        # than that in the tax unit file because it's possible
+        # dependents have income, but aren't filing so it's not counted
+        # we're only raising an error if the difference between the two
+        # values is greater than .1 because for some reason
+        # interest income would raise an error even though the totals
+        # appeared to be the same
+        # error = cps_sum < tc_sum and abs(cps_sum - tc_sum) > .1
```
MaxGhenis, line 93: why commented? also looks like the preceding comment refers to this, should that be deleted?

cps_stage2/stage2.py
```diff
@@ -1,70 +1,84 @@
...
+    stage_1_factors = pd.read_csv(
+        STAGE_1_PATH, index_col=0
+    )
+    stage_2_targets = pd.read_csv(
+        STAGE_2_PATH, index_col=0
+    )
     # DataFrame for holding each year's weights
     weights = pd.DataFrame()
+    print("Solving for 2014")
     weights['WT2014'] = solve_lp_for_year(cps, stage_1_factors,
```
MaxGhenis, line 28: a loop here would be more concise

tests/test_data.py
```diff
@@ -182,44 +193,43 @@ def check_cps_benefits(data):
     """
     bnames = ['mcare', 'mcaid', 'ssi', 'snap', 'wic',
               'tanf', 'housing', 'vet', 'other']
-    expect_minben = 0
     # specify expected benefit statistics in CPS data
     expect_ben_stat = dict()
```
MaxGhenis, line 197: uppercase for constant (ideally a CSV)

tests/test_data.py
```diff
@@ -182,44 +193,43 @@ def check_cps_benefits(data):
     """
     bnames = ['mcare', 'mcaid', 'ssi', 'snap', 'wic',
```
MaxGhenis, line 194: uppercase constant

tests/test_weights.py
```diff
@@ -50,7 +50,7 @@ def test_weights(kind, cps_weights, puf_weights,
                                         weights[col].max(), max_weight))
     # test sum of weights (in millions) for each year
     min_weight_sum = 149
```
MaxGhenis, line 52: uppercase (also max_weight_sum)

cps_data/pycps/impute.py
andersonfrailey: Good catch! Should be uncommented.

cps_data/pycps/impute.py
andersonfrailey: All of the hard coded numbers come from John and can be found in [this](https://github.com/PSLmodels/taxdata/blob/master/cps_data/State%20Database%20-%202015/Production%20File/Imputations/Impute3TobitV4.sas) file. I carried them over to be consistent with what we've been doing for imputations up to this point. Agreed that it's less than ideal. I'd like for one of the next TaxData projects to be revisiting all of the imputations we do on the CPS to make them more transparent, reproducible, easier to update, etc. Getting rid of all the hard coded numbers would be part of that work.

cps_data/pycps/targeting.py
andersonfrailey: It incorrectly specified which target to use for `CGAGIX`. Should've been deleted

cps_data/pycps/validation.py
andersonfrailey: I was re-factoring the validation scripts and didn't want to delete it just in case things didn't work. Forgot to go back and delete it later. Gone now 👍🏾

cps_data/pycps/validation.py
andersonfrailey: Yep. This test is taken care of on the next line. Deleting

andersonfrailey:
Thanks for the review, @MaxGhenis! Pushing updates in response shortly

andersonfrailey:
Pushed everything new up

andersonfrailey:
Are there any more comments on this or can I go ahead and merge?

@MattHJensen @MaxGhenis @Peter-Metz @hdoupe

MattHJensen:
@andersonfrailey, this is very exciting!  

Would you mind (re) summarizing your work when you have an opportunity? (The top summary comment is somewhat out of date now thanks to all of your updates and Max's feedback!)

cps_data/pycps/finalprep.py
```diff
@@ -0,0 +1,285 @@
...
+    # assert that we don't lose any of the variable
+    tot = (var*weight).sum()
+    m = f'{tot}:::{goal_total}'
+
+    assert np.allclose(np.array(goal_total), np.array(tot)), m
+
+    return var
+
+
+def adjust(data, targets):
```
MattHJensen, line 164: @andersonfrailey, it doesn't appear that `adjust()` or `adjust_helper()` is taking into account the filer/non-filer distinction when creating the adjustment factors. 

`target()` seems very similar to `adjust,` just focussed on a wider variety of state-based adjustments. (@andersonfrailey, if I am missing anything in that assumption I'd love a correction). 

`target()` calculates the adjustment factor as the difference between the filer population's sum and the SOI target, and then it applies the adjustment factor across both filers and non-filers. This generally makes sense to me (and seems like a fine approach for this PR since it doesn't seem to be a departure from the SAS code), but I want to think more about whether it ends up skewing the filer/non-filer count one way or the other.

andersonfrailey: @MattHJensen, in addition to focusing on a wider variety of state-based adjustments, `target()` actually changes the total value of the targeted variables while `adjust()` only changes the distribution of the targeted variable in the data. We don't make a distinction between filers and non-filers in `adjust()` because we aren't changing the totals.

Neither of these would change the unweighted total of filers/non-filers because we make that distinction before any adjustments are made, but they could affect weighted sums after the stage 2 adjustments.

andersonfrailey:
>Would you mind (re) summarizing your work when you have an opportunity?

Happy to!

I'm working on much more detailed documentation of what's happening here, but the gist is that this PR ports the code used to create `cps.csv.gz` from SAS to Python. The general steps we take are:
1. Create tax units from the 2013, 2014, 2015 CPS March supplement files. The CPS data information on marital status, if/by whom an individual is claimed as a dependent, and filing status. We use that to form tax units within each household. The tax units from each of these individual files are combined into one dataset.
2. We split up certain income data we have in the CPS to match how it's classified in the IRS Public Use File. For example, dividend income reported in the CPS lumps qualified and ordinary dividends in one variable. We split what's reported in the CPS based on the ratio of the two in the 2011 PUF.
3. We target some income variables so that the weighted totals among those determined to be tax filers in the CPS match what the [IRS reports](https://www.irs.gov/statistics/soi-tax-stats-state-data-at-a-glance-fy-2014).
4. We impute a number of itemized deductions that are not found in the CPS to round out the data in the file.
4. As we're cleaning up the file, we target the distribution of interest, dividend, and business income to match those found in IRS data.

All of the tax units also have benefit program data imputed from C-TAM.

A couple of additional files are added by the PR unrelated to the logic behind creating tax units. `validation.py` is used to validate the income totals and person counts in each tax unit. `transform_sas.py` and `template.txt` are used to translate the SAS code provided by NBER to convert the CPS files from a .dat format to something useable.

cps_data/pycps/finalprep.py
MattHJensen: Thanks Anderson. Resolving.

MattHJensen:
Thanks a lot for the overview Anderson. I am ever so happy to be leaving SAS behind and entering a new realm of continuous improvement in Python thanks to this work!

andersonfrailey:
>I am ever so happy to be leaving SAS behind and entering a new realm of continuous improvement in Python

You and me both! I'm going to tag the current version of taxdata just in case we need to revisit the SAS code for whatever reason, and then merge this PR at the end of the day if there are no objections.

Issue #323, opened 2019-07-10
Weight and age CPS tax units in the same household together


MaxGhenis:
## Motivation

Andrew Yang's Freedom Dividend is a UBI restricted to adults, plus some tax increases. Since some children file taxes separately from their parents (https://github.com/PSLmodels/taxdata/issues/321), this causes some tax units to be worse off, since they experience the tax increase without the UBI ([analysis](https://medium.com/ubicenter/why-some-low-income-people-come-out-behind-in-andrew-yangs-freedom-dividend-c95d0f1cc433)). These are low-income tax units, so it makes the plan look overly regressive. This doesn't reflect reality, however, since nearly all share resources with household members that receive the UBI. 

Aggregating tax units back to households would address the problem, but this is invalid (https://github.com/PSLmodels/Tax-Calculator/issues/1961), because tax units in the same household have different weights, and they're aged separately.

Others have reaggregated tax units to household or SPM unit and used the original household/SPM unit weight (just adjusting the total tax liability), but this is limited to the most recent CPS tax unit year, and pending the new Python scripts, also relies on separate tax unit creation code for anything newer than 2015. This won't be as big an issue when 2018 CPS tax units are available, but someone could still want to project into the future.

Reaggregation aside, it seems intuitively inaccurate to assign different weights to tax units within the same household.

## Suggestion

Would it be possible to add constraints to the weighting and aging routines to ensure that tax units within the same household have the same weight?

MattHJensen:
This is also discussed at the end of page 142: https://www.irs.gov/pub/irs-soi/06foertsch.pdf

MattHJensen:
This report may also be relevant at pg 4: 

https://www.cbo.gov/system/files/2019-12/55941-CBO-Household-Income.pdf

![image](https://user-images.githubusercontent.com/8301092/85022272-0ef2b880-b141-11ea-9732-05791c581f95.png)

PR #324, opened 2019-07-11
Refactor solve_lp_for_year.py

MaxGhenis:
*Not ready to merge*

In filing #323 I looked through this file, and took a stab at streamlining the code. Changes include:
1. Using a `wage_bin` function to create wage bins.
2. Using a loop to assign `lhs_vars` dictionary elements.*
3. Using a loop to assign factor variables.
4. Using an `add_target` function to assign `rhs_vars` dictionary elements.*
5. More explicitly adjust `model_vars` for year 2014.
6. Rename `LP` to `lp` since it's not a constant.
7. Other minor changes like comments.

\* These pieces won't currently work since they attempt to access `globals()` from within a function (`locals()` also won't work). Per https://stackoverflow.com/q/56983782/1840471, there's no good way to access these in-between-scoped variables, and this is really a sign that something could be done better. It seems to me that sticking with the dictionary elements rather than creating new variables would be a clean alternative.

I haven't tested this code yet. Would the right way be to run `make cps-files`?

### DIFF
--- a/cps_stage2/solve_lp_for_year.py
+++ b/cps_stage2/solve_lp_for_year.py
@@ -48,59 +48,39 @@ def solve_lp_for_year(data, factors, targets, year, tol):
     ucomp = data.UCOMP * s006
 
     # wage distribution
-    wage1 = np.where((data.e00100 <= 10000), data.WAS, 0) * s006
-    wage2 = np.where((data.e00100 > 10000) & (data.e00100 <= 20000),
-                     data.WAS, 0) * s006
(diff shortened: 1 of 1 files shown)


andersonfrailey:
Thanks for working on this, @MaxGhenis! These changes look good so far.

>I haven't tested this code yet. Would the right way be to run make cps-files?

Yes. Once you run this it'll create a new weights file and you can see if the results have changed at all. Fair warning, this will take a few hours to run.

MaxGhenis:
Sorry I've left this open so long. I'll resolve conflicts, boil it down to the basics, and test it, once #343 is merged.

Issue #311, opened 2019-04-10
Add count of children under age 6


MaxGhenis:
Two major tax credit bills split out children under age 6:
* [American Family Act](https://www.vox.com/future-perfect/2019/3/6/18249290/child-poverty-american-family-act-sherrod-brown-michael-bennet) (Bennet-Brown child allowance)
* [Working Families Tax Relief Act](https://www.vox.com/future-perfect/2019/4/10/18302183/tax-cut-democrats-earned-income-tax-credit-child-allowance)

Currently the `nu05` field and `_CTC_c_under5_bonus` parameter threshold at age 5, so it'd be helpful to have parallel versions of these thresholded at age 6.

martinholmer:
Why not **replace** `nu05` with `nu06`?  The only place in Tax-Calculator code that `nu05` is used is in the `taxcalc/reforms/Clinton2016.json` file.

@MaxGhenis @andersonfrailey @MattHJensen

MaxGhenis:
SGTM

andersonfrailey:
I'm also on board with @martinholmer's suggestion. I can get that done for the PUF this week. However we still can't run the SAS scripts to recreate the CPS file. I have an idea to use the record identification variables we've left on the CPS file to link it with the raw CPS files to work out counting variables like this would be, but that'll take a little more time to develop and test.

martinholmer:
@andersonfrailey said 13 days ago in taxdata issue #311:
> I'm also on board with @martinholmer's suggestion. I can get that done for the PUF this week. However we still can't run the SAS scripts to recreate the CPS file. I have an idea to use the record identification variables we've left on the CPS file to link it with the raw CPS files to work out counting variables like this would be, but that'll take a little more time to develop and test.

What's the schedule for getting this done?  Tax-Calculator users are waiting for this enhancement as you can see from the discussion of [Tax-Calculator issue 2294](https://github.com/PSLmodels/Tax-Calculator/issues/2294).

@MattHJensen

andersonfrailey:
@martinholmer, I could get a new PUF out with the change from `nu05` to `nu06` this week. However, there are a few things to consider with making the change in the CPS. As I mentioned in the last Tax-Calculator maintainers meeting, I want to work on creating a new CPS file that is built with Python scripts so that we're no longer constrained by the SAS code we currently use. If we wait to release a new CPS file until that's ready, it will be at least a few weeks.

[...]

PR #329, opened 2019-11-02
Add nu06 Variable to PUF

andersonfrailey:
This PR changes the `nu05` variable to `nu06` in `puf.csv`. As of yet, we can't make this update in the CPS file, but as @MattHJensen pointed out in issue #311, there are users who would like to see this added to the PUF as a first step.

Before merging, I think we should establish what to do with the `nu05` variable in the CPS file. I'm inclined to either a) remove it or b) actually keep it in the PUF so that we will have `nu05` and `nu06`. I'm open to other's suggestions on how to move forward from here.

Also, I found that when I created the new PUF file, the `e19200` variable sum increased by 1. This is probably just rounding error, but I think it's still odd that this has happened.


MattHJensen:
> Before merging, I think we should establish what to do with the nu05 variable in the CPS file. I'm inclined to either a) remove it or b) actually keep it in the PUF so that we will have nu05 and nu06. I'm open to other's suggestions on how to move forward from here.

I'm inclined to keep both in puf.csv until we add `nu06` to cps.csv, at which time we drop `nu05` from both files, cps.csv and puf.csv. I don't feel strongly about this though if someone else has an alternative preference 

(cc @MaxGhenis).

MaxGhenis:
I'm not aware of major policies using `nu05` so up to you how long to keep it once `nu06` is added (replacing would be OK).

MattHJensen:
@andersonfrailey, are you happy with this approach: 

> I'm inclined to keep both [`nu05` and `nu06`] in puf.csv until we add `nu06` to cps.csv, at which time we drop nu05 from both files, cps.csv and puf.csv.

andersonfrailey:
@MattHJensen Yep this is good with me. I'll update the PR accordingly.

andersonfrailey:
Added nu05 back to the PUF in the latest commit

MattHJensen:
@andersonfrailey, could you please update me on the status of this PR? I apologize if it was ready to be merged before and requires an update now.

andersonfrailey:
@MattHJensen, ready to go! There doesn't seem to be any conflicts. We still can't add `nu06` to the CPS until PR #314 is merged, but I think that one is ready to go as well.

andersonfrailey:
Now that #314 is merged, this will be ready to go when I get the merge conflicts resolved

andersonfrailey:
Fixed up the merge conflicts. Totally removed `nu05`, tests are passing. This is ready to go
@MattHJensen

andersonfrailey:
Doesn't look like there are any more comments so merging

PR #339, opened 2020-07-03
Add 2016, 2017, 2018 CPS Support

andersonfrailey:
This PR modifies the CPS scripts to support using the 2016, 2017, and 2018 March supplements and makes it generally possible to not use years where we don't have C-TAM imputations yet. Some things to note:

* Values for Medicaid, Medicare, WIC, and (for 2016 and beyond) housing benefits are not included in the CPS so the final datasets produced without C-TAM imputations do not either. As such, when we distribute `other_ben`, the only variable we do it by is `SSI`.
* I modified the TaxUnit creation process to work without C-TAM benefits. This did not cause any change in the final file produced when using the 2013, 2014, and 2015 supplements with C-TAM imputations.
* While working on this, I noticed that in 2016 and 2017 there is no business income (`e00900`) in the highest income bin we use for the adjustments in `finalprep.py`. That seems odd to me and worth investigating at some point. But the result of this is our adjustment method fails to keep the weighted sum of `e00900` unchanged. My solution for now is to simply reverse the adjustment. Eventually we should look into changing the methodology or determining why we see no business income in those bins.
* I didn't change the stage 2 scripts at all since we're sticking with 2013-15 as the main files. I'll add some documentation for others on how to update the stage 2 scripts if they want to use a different year.


andersonfrailey:
If there are no objections, I'll merge this in on Monday.

andersonfrailey:
Actually, given the discussion in issue #342, I'm reworking how we process the raw CPS files. Once that is done, this PR will be unneeded. I'll leave it up for now and close it when I get the new PR up.

PR #344, opened 2020-07-16
Makefile update

chusloj:
This PR just makes a small update update to the `Makefile` to change the path `cps_data/cps_raw.csv.gz` to `cps_data/pycps/cps_raw.csv.gz` within the files needed for `cps_stage2/cps_weights.csv.gz`.


chusloj:
Changes in this PR are addressed in #343. This PR can be closed.

Issue #348, opened 2020-07-30
Rewrite Stage 2 in Julia


chusloj:
There is a replicability issue using the `CVXOPT` solver in Python to calculate the PUF and CPS weights - the `md5` hash for the weight file changes every time the solver is run. Because most of the commonly used LP solvers do not have clean APIs for Python, I'd like to rewrite the `stage2` and `solve_lp_for_year` scripts in Julia. The language has a clean optimization & modeling interface called [`JuMP`](https://jump.dev) that can be used for any LP model that has a Julia implementation.

There are a few reasons why using Julia would be advantageous to Python for the solver stage:

1. Julia has a `pandas` interface, so refactoring the data processing portions of the code shouldn't be too time consuming.
2. Using `JuMP`, any new solver can be used with the same code because `JuMP` is model-agnostic.
3. Julia is a compiled language, so it might have a speed advantage over Python.

Coding the Julia version of the code should take substantially less time than finding the `md5` replicability bug.

andersonfrailey:
Just to add a little more information about the replicability issue, @chusloj and I have been going back and forth about how the `cps_weights.csv.gz` file changes in PR #343. As part of our investigation into why that was happening we each re-ran the CPS stage 2 scripts multiple times and could never get the same `md5` hash. I even compared the hashes between multiple weights files I created on my machine off the master branch and each one was different.

@chusloj did some comparisons and it seems like the differences between files are very small, but they're different nonetheless. Given all of this and the fact that the `pulp` solver we currently use for the PUF takes six to eight hours to run, I'm in favor of converting all of the stage 2 scripts to Julia and if the results look good making that our permanent solution.

donboyd5:
I have been running `stage2.py` on the puf (although still need to understand #352), which led me to some thoughts about your rewrite of stage 2. I think the comments (if on target) are relevant regardless of whether written in Julia or python.

As I watched `stage2.py` solve, I saw that each new year seemed to take more iterations than the year before, generally resulted in a higher objective function value, and often took a bit more time although that was not always true.

This led me to look at the code in `stage2.py` and `solve_lp_for_year.py`. If I read it properly (not 100% certain of this), I think the penalty function in all cases is based on how different the new weights are from the 2011 weights -- that is, the 2012 penalty is based on difference between 2011 weights and solved-for 2012 weights, ..., and the 2030 penalty is based on the difference between 2011 weights and solved-for 2030 weights.

[...]

PR #351, opened 2020-08-03
Update stage 2 to Julia

chusloj:
Work in Progress.

This PR addresses #348 . This will replace both the CPS and PUF stage2 Python files with Julia files and change `README` files accordingly.


MaxGhenis:
@chusloj FYI, I refactored some of the Python code behind this in #324 to avoid code duplication. Happy to do that after you're done, though I might ask you for some tips as I haven't worked with Julia before, or if it's easy for you to do while you're in here feel free.

Also some of the multi-line statements look like they need some extra indentation to align.

chusloj:
The code for the Julia solver update is finished. Overview:

- The `solve_lp_for_year` file has been split into a `dataprep.py` file and a `solver.jl` file
- The `stage2.py` file has been refactored to remove redundant code

The build is not passing due to `test_data.py` because `puf_agg_actual.txt` and `puf_agg_expected.txt` differ.

@andersonfrailey

andersonfrailey:
Just to put this in there before I forget, we need to install three Julia packages to run the new stage 2 scripts: `Cbc, JuMP, NPZ`. I haven't checked to see if there's a way to specify this in the `enviornment.yml` file. If not we'll just need to specify the need to install them in the docs.

chusloj:
A Julia environment was just created and added in the new `.toml` file. The structure is similar to `.yml` files used in `conda`, but Julia's package manager does not have an option for activating a package at the command line, so the environment is being activated when `solver.jl` is run.

chusloj:
After comparing Tax-Calculator projections from the PUF and CPS weights before and after the Julia solver was implemented, results show that projections don't change.

If this seems irregular, note that the solver Julia is using right now ([`Cbc`](https://github.com/jump-dev/Cbc.jl)) is the same solver that `PuLP` was using before Julia & `JuMP` were implemented.

andersonfrailey:
Thanks for working on this, @chusloj. I'm going to leave this PR open for a couple days to allow feedback from others.

PR #359, opened 2020-08-22
Update to 2020 OASI Report

andersonfrailey:
This PR updates our CBO baseline calculations to use the 2020 OASI Annual Trustees Report. Based on the change log, it also appears that the BLS has updated their CPI-M numbers.


andersonfrailey:
If there are no objections, I'll merge this at the end of today.

Issue #360, opened 2020-08-24
Adopting Black Code Formatting


andersonfrailey:
I'd like to propose we add a [black](https://black.readthedocs.io/en/stable/) pre-commit hook to TaxData to handle code formatting. It'd be nice to not have to worry about that when reviewing PRs. If there are no objections, I can open a PR to add it.

hdoupe:
I've really enjoyed using Black. Once you have it hooked into your editor to run every time you save, you'll never look back. 

Downside is that you pretty much have to run it on your entire repo which will ruin a lot of git history. However, maybe there is a way around this--I just haven't seen it yet.

andersonfrailey:
Throwing off the git history is a concern of mine as well, @hdoupe. But better to do it now than after we do all of the refactoring that's been discussed in #340 and #357.

hdoupe:
ParamTools uses Flake8 in its pre-commit [config](https://github.com/PSLmodels/ParamTools/blob/master/.pre-commit-config.yaml), too. It's a little bit more of an undertaking to fix all of the initial errors, especially for a large project like taxdata. But as annoying as it can be, it does catch a bunch of random stuff like unused imports and unused variables.

andersonfrailey:
That would be good to have. Copying the ParamTools pre-commit config might be the move here.

MaxGhenis:
Yep I just started using Black and am going through the initial cleanup for `microdf` - it's a pain at first but I'm glad I'm doing it. Seems like a potential "SHOULD" criterion for PSL acceptance too.

Were you thinking of using https://github.com/lgeiger/black-action? I'm using https://github.com/WhyNotHugo/python-linting which I think does Black and more.

andersonfrailey:
@MaxGhenis I was thinking of using https://github.com/lgeiger/black-action. I hadn't heard of https://github.com/WhyNotHugo/python-linting before but it looks interesting. I'll review both a little more.

andersonfrailey:
I think copying the ParamTools pre-commit config will be a good solution for taxdata. If I understand how it works correctly, that'll catch everything and run black before code is even pushed to GitHub. I'll open a PR later today.

PR #362, opened 2020-08-31
Use Pre-commit hooks

andersonfrailey:
This PR adds pre-commit hooks to the repo as discussed in #360. As expected, there are a ton of "changes" that amount to just formatting changes. This does require developers to run `pre-commit install` for them to work correctly. I've added that to the install instructions.


andersonfrailey:
As expected, these updates have no effect on any of the files taxdata creates. If there are no objections, I'll merge this tomorrow.

PR #387, opened 2021-05-24
Fix zero weights issue

andersonfrailey:
This PR addresses #385. All we needed to do was rename the weights for non-filers from `s006` to `matched_weight` so that they were properly appended to the filers. Thanks to @donboyd5 for flagging this issue.

A few other notes:

For some reason, the unweighted sum of `e00200` changes slightly. It's unclear why because nothing I changed should have affected this. I want to try and get to the bottom of this before merging this PR.

I removed `Project.toml` and `Manifest.toml`. These two files were used by Julia to specify packages and version numbers used in our stage 2 process. However I don't think they were specified properly because I repeatedly got errors about packages not having the proper dependencies listed in `Manifest.toml`. I think it would be easier to just include in the documentation instructions for installing the proper packages. I'm interesting in hearing other people's thought here or if they've run into similar troubles.

Issue #386, opened 2021-05-23
Skipping Years During Stage 2


andersonfrailey:
As mentioned in #381 and other issues, running our stage 2 scripts takes forever. Beyond changing solvers, another change I want to make to speed this process up is to only run the LP model for years that have seen their targets change. For example, in the PUF the targets for 2012 to 2017 have not and will not change because the CBO won't be releasing any new data and we won't be changing out SOI estimates. This means the final weights also will not change and solving for them is really just a waste of time and resources.

I'd like to propose simply skipping these years by adding a check to the stage 2 scripts that compares the targets checked into this repo for a given year to those on the local machine of whoever is running the scripts and skipping the solver for that year if they're the same. I also want to add a file to the repo that contains the MD5 check-sum for `cps-matched-puf.csv` and `cps.csv.gz`. These will be compared to the local files as well so that if there is a change to either we'll create new weights for all years. For the years that are skipped, we'll just use the weights that have already been created.

A minor downside to this is it would kind of lock us into a solver because the checks wouldn't been looking to see if we're using a new solver. We can overcome this by adding an option to force the model to solve for all years. 

I believe these checks should prevent skipping years that we actually need, but if anyone can think of additional checks I should impose please let me know.

cc @donboyd5 @MattHJensen

donboyd5:
@andersonfrailey, I really like this idea.

PR #390, opened 2021-05-25
Skip Stage 2 years

andersonfrailey:
This PR implements the suggestion in issue #386. It's still a work in progress, but here's a high level summary.

I created hashes for all of the stage 2 scripts and datasets and saved them in `datahashes.json`. The stage 2 scripts then compare these hashes, the targets, and the factors before running the LP model. If all of these are equal for a given year, the solver is skipped and the weights we already have are used again. I believe that these checks will be sufficient to catch any changes that would effect the final weights. All of the initial hash values are created by `inithash.py`. If the input files are changed, they'll need to be updated after creating the new weights.

For some reason, the PUF weights are being changed right now even after skipping the solver. I need to get to the bottom of that bug before this PR is ready to merge. I also think it should be merged after we switch solvers and after #387.


martinholmer:
@andersonfrailey, Thanks for clarifying the rationale for pull request #390.

Issue #381, opened 2021-05-19
Possible 85-95% reduction in stage 2 reweighting time using drop-in-replacement solver in solver.jl


donboyd5:
I don't know if others have found reweighting to be as slow as I have, but stage 2 reweighing for CPS took more than 14 hours on an older but still reasonably fast CPU. That's more time than I can tie a computer up for ordinarily.

As some of us have discussed, this really shouldn't take more than half a minute or so a year with an NLP solver and slight problem restructuring.

But in this case I was looking for a pure LP solution that required no restructuring.

After looking over the list of [available JuMP LP solvers](https://jump.dev/JuMP.jl/v0.21.1/installation/), I decided to investigate Tulip ([here](https://github.com/ds4dm/Tulip.jl) and [here](https://arxiv.org/pdf/2006.08814.pdf)).

After installing, it can be dropped into [solver.jl](https://github.com/PSLmodels/taxdata/blob/master/cps_stage2/solver.jl) with just the following code changes:

```
# Add Tulip to the packages used
using JuMP, Cbc, NPZ, Tulip  # drop Cbc if done testing
...

# Replace these next 2 lines:
# model = Model(Cbc.Optimizer)
# set_optimizer_attribute(model, "logLevel", 1)

# With these 3 lines:
model = Model(Tulip.Optimizer)  # djb
set_optimizer_attribute(model, "OutputLevel", 1)  # 0=disable output (default), 1=show iterations	
set_optimizer_attribute(model, "IPM_IterationsLimit", 500)  # default 100
...

# Add these 2 lines to see final objective function:
st = termination_status(model)
println("Termination status: $st")
...

```

Here are summary results. As you can see Cbc takes a quarter-million iterations. For Tulip, I raised the default limit on number of iterations from 100 to 500 so that the solution values are almost identical to those from Cbc for 2014:

**Objective function:**
Cbc (taxdata):           34,003.997
Tulip (500 iter)          34,004.8231
Tulip (100 iter)          34,027.7230

**Iterations:**
Cbc (taxdata):            253,373
Tulip (500 iter)                  500
Tulip (100 iter)                  100

**Solution time (minutes):**
Cbc (taxdata):           48.8
Tulip (500 iter)             6.4
Tulip (100 iter)             1.3

**Quantiles (0, .1, .25, .5, .75, .9, 1):**
Cbc:                       [0.3,              1. ,                 1. ,                   1. ,        1. ,                  1.7]
[...]

PR #393, opened 2021-06-10
Switch to Tulip Solver

andersonfrailey:
This PR is in response to issue #381. It switches us over to the Tulip solver for the stage 2 process. When running the scripts this morning it took about an hour and a half to create weights for both the PUF and CPS, compared to the 24+ hours it took last time I created the weights with our current solver. I'll post results comparisons in a followup comment.

cc @donboyd5


andersonfrailey:
Relevant changes to the calculations can be found [here](https://github.com/PSLmodels/taxdata/blob/06b169099055f152b8b792661a398706110759c6/history/reports/taxdata_report_2021-06-10.pdf). There are no major changes in output when compared to taxcalc 3.0.0.

donboyd5:
Thanks @andersonfrailey. It looks to me like this is based on the early version of notes I made regarding Tulip. However, if you work through the full post of #381 you will find two major additional improvements:
- scaling of the constraints, which makes the problem more stable numerically and allows it to reach optimality easily
- replace an inefficient approach that was being used in taxdata whereby ~250k constraints were being added unnecessarily, with an approach that achieved the same goal without those unnecessary constraints by instead placing bounds on the variables

If you scroll toward the bottom of #381 you will find a drop-in replacement solver.jl function that implements all of the changes:

1. Tulip instead of Cbc
2. scaling of constraints
3. replace unnecessary constraints with variable bounds

If you use this code (it should be easy to replace what you have now with this) I think you will find that it (a) reaches optimality easily, (b) it is MUCH faster still, and (c) it is more stable numerically.

andersonfrailey:
Last commit makes the changes suggested by @donboyd5

andersonfrailey:
If there are no more comments, I'll merge this at the end of the weekend.

MattHJensen:
Out of curiosity, what was runtime after the latest round of changes?

andersonfrailey:
@MattHJensen running the full make script takes about 30 minutes now!

PR #394, opened 2021-06-16
Update CBO baseline

andersonfrailey:
This PR updates taxdata to the February 2021 CBO economic projections. I'll post a summary of how it affects Tax-Calculator projections in my next comment.  One note on testing.  I had to drastically change the max and minimum growth factor because of the spike in unemployment insurance due to the COVID-19 response.

Unless someone else has an idea for additional work, I believe this should be the last PR before the next taxdata release.

cc @MattHJensen @jdebacker


andersonfrailey:
Updated projections can be seen [here](https://github.com/PSLmodels/taxdata/blob/c7e907265e38748aadfd34543dd3954c750d96e3/history/reports/taxdata_report_2021-06-16.pdf).

andersonfrailey:
For future reference, I added instructions to `updatecbo.py` saying to input the February 2021 unemployment numbers manually because CBO changed the file format for that release, but the new file seems very pandemic-year specific so, in my opinion, it's not worth changing the code base to fit this new file. If future CBO updates keep the same format, then we can update our code accordingly.

jdebacker:
LGTM - nice work @andersonfrailey!

PR #402, opened 2021-08-12
Add automatic download for CPS files

jdebacker:
This PR added and automatic download for the March CPS file.

@andersonfrailey I will also add a download for the `cps_raw.csv.gz`, but I don't know anything about the provenance of this file.

### DIFF
--- a/taxdata/cps/create.py
+++ b/taxdata/cps/create.py
@@ -1,5 +1,9 @@
 import pandas as pd
 import pickle
+import requests
+import zipfile
+import io
+import os
 from . import validation
 from pathlib import Path
 from tqdm import tqdm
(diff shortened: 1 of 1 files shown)


jdebacker:
@andersonfrailey pinging - let me know how this looks.

andersonfrailey:
This looks great, @jdebacker! Thanks for working on it. Seems ready to merge from my perspective. Are there any final changes you want to make before I merge?

jdebacker:
@andersonfrailey Thanks for taking a look.  Yes, I'd like to add a download of the `cps_raw.csv.gz`.  Where exactly is that from?

andersonfrailey:
@jdebacker there's nowhere to download that file. It gets created as you're running the CPS scripts and we've left it out of the repo to keep the size down.

jdebacker:
@andersonfrailey Got it.  Ok - then this PR is ready to merge.

PR #411, opened 2022-01-10
Fix PUF SOI estimates

### LINKED ISSUE #399: $121b of wages of millionaires dropped from puf stage 2 targets??
Of possible interest to @andersonfrailey @MattHJensen @rickecon @jdebacker:

Been trying to figure out why some of my numbers for 2017 differ substantially from official TaxData &  Tax-Calculator so I looked at wage data for 2017. As noted in the title, I think TaxData is dropping $121 billion of wages of millionaires, which would help explain why I get such very different distributions of income and other items vs. TaxData. Here are key facts and inferences:

- I can see that taxdata puf stage 2 targets for wages by income range appear to come from the same source as mine: the IRS file "17in14ar.xls" (17=2017, in=individual income tax, 14=Table 1.4, ar=all returns -- [landing page](https://www.irs.gov/downloads/irs-soi?sort=asc&order=Ng%C3%A0y&page=413)).

- For example, here is a doctored screenshot of the [stage 2 targets](https://github.com/PSLmodels/taxdata/blob/master/puf_stage1/Stage_II_targets.csv) that just shows wages, just for 2015-2017:

image

- And here is a screenshot of the wage portion of IRS Table 1:
image

- If you look at the first 2017 wage target labeled as "Wages and Salaries: Zero or Less" it matches the IRS "no adjusted gross income" value of $20.869 billion. So far so good, although the target label implies that it is based on wage bins but the IRS data are based on AGI bins.

- (Before going further, let me note that this raises a side issue: the IRS data list values by 2017 AGI range, for tax filers. When I retarget the puf, I first calculate AGI for the year in question, then determine filing status, and apply targets to filers in corresponding AGI bins. Unless TaxData calculates AGI before targeting - I don't think it does but I need to read the code - it must be targeting for all records (not just filers) and more importantly, based on some different definition of AGI. I seem to recall it might use the base year of 2011, which would be far away in time from 2017. If that's how it's done, it would seem to result in wage distributions for 2017, after AGI is calculated, that are too light on the high end if I am thinking about this properly. But that's not the question I'm writing about here.)

- The second target ($1 < $10k) of $86.507 billion matches the sum of the next 2 IRS AGI ranges, which cover $1-10k, so that looks good. And the next few ranges that I looked at also look good. For example, the $500k-1m range target and corresponding IRS value both are $379.376 billion.

- But when we get to millionaires, we have a problem, as far as I can tell. TaxData shows a target of $367.732 billion. Let me zoom in on that here:

image

- However, the IRS sum for millionaires is $489.1 billion, a full $121 billion greater. The screenshot below zooms in on the IRS data and calculates the sum for millionaires, and the sum for those at $1.5+ million, leaving out the $1-1.5 million group.

image

- As you can see, the puf targets appear to have dropped the $121 billion in wages for those with 2017 AGI of $1-1.5 million. A quick scan across the millionaire row suggests this happened in all years.

- Assuming I did my work correctly, this means that puf.csv as reweighted and grown generally will have too little wages overall, too little wages for millionaires, and some distributional oddities as records are reweighted in an effort to hit wage targets that are too low for millionaires.

If I made a mistake in examining this I would much appreciate an early alert.

andersonfrailey:
This PR addresses issue #399.

The `updatesoi.py` file is used to automatically update our SOI estimates, but the range of indicies used to add up total wages for those with AGI greater than $1 million was off and excluded some of the data, as @donboyd5 figured out. This PR fixes that and adds an additional check to  `updatesoi.py` to prevent an issue like this from going undetected in the future.

The bug affected our SOI estimates for 2015-2017. They have all been fixed in this PR.


andersonfrailey:
This PR is just about done, but with the changes there's a big increase in tax liability for 2030 and 2031 that I can't explain. I've attached a table comparing the tax liabilities that were found with taxcalc 3.2.1.
Screen Shot 2022-01-10 at 2 58 31 PM

donboyd5:
Would it be hard to construct a table for two (or more) of the years that
gives total wages by an income classifier (e.g., AGI range) pre-PR and with
PR for (a) the year before the first surprising year (i.e., 2029), and for
the first surprising year (e.g., 2030)?

For example, tables a and b would have stubs such as:

income range        wages pre-PR    wages PR   change
<= $0
...
$1m-10m
$10m+
sum

We'd then see (1) which income ranges are causing the problem, and (2) how
those income ranges changed between 2029 and 2030.  Of course it could be
something else, but this seems like the most likely suspect.

If we verify that the issue is caused by changes in wages, and we figure
out which income ranges are at work (probably the top 2 or 3), it would
then make sense to look at growfactors moving from 2029 and 2030, and at
wage targets for 2029 and 2030. My guess (uniformed) is that there is
something surprising about the targets in 2029 but it could be growfactors.
Of course it could be something else entirely, such as # returns with wages
(again the same sort of breakdowns with #s would be helpful), or even
something out of left field, but I'd suggest looking at wages first.

Don



My guess is that that would tell us a lot about what happened, and then
we'd have to look at growfactors and target weights

andersonfrailey:
There are some big changes in total wages in the later years, both in terms of raw and weighted wages. But where in the distribution they occur is different between the weighted and unweighted totals. The charts below are the differences in wages in different groups for 2029-2031.

Unweighted:
Screen Shot 2022-01-11 at 11 07 43 AM

Weighted:
Screen Shot 2022-01-10 at 5 02 11 PM

This makes it look like there's an issue with growth rates, but the growth rates look fine. That would leave the linear programming model, but I don't know why any issues there would only affect the last two years

andersonfrailey:
Deleted my last comment, Realized there was an issue with how I was looking at the data when I was making those charts

jdebacker:
@andersonfrailey Is this PR ready for review?  Last comment made that unclear.  Also, could be helpful to produce tables Don suggested to aid in review.

andersonfrailey:
@jdebacker, I haven't been able to fix the issue with tax revenues jumping up in the last few years yet, but I'd definitely be open to others seeing if they get the same result when they run the changes in this PR. Spring break starts tomorrow so I should be able to work on those tables Don suggested this week!

jdebacker:
Great - thanks for the update!

MattHJensen:
@andersonfrailey, I ran `make puf-files` with this PR and ran into several ITERATION_LIMIT and INFEASIBLE terminations during stage 2. Is this to be expected? Terminal output in [this gist]( https://gist.github.com/MattHJensen/8040d1fb7867c70f03e001aff8aea8d3).

I'm going to push forward to replicate your revenue table and then create Don's suggested tables in the next few days, but thought I should check in on this. Thanks!

donboyd5:
A few thoughts:

**2016:**
image

- it hit the iteration limit of 100
- the targets were satisfied (the percentile values for ratio of calculated to desired targets are all approximately 1)
- the ratios of new weights to old weights ranged from 0.69 to 1.86
- but the computer code says the tolerance around the weights was 0.30 in 2016, meaning we'd want the ratio to range from 0.70 to 1.30; thus, it was only able to hit the targets by making some weights larger than it was told to make them

image
- I doubt that raising the iteration limit would solve this
- you might find this solution acceptable and might consider increasing the tolerance around the weights (e.g., +/- 0.90, or make it asymmetric (with a coding change))
- or you might examine the targets - they may be very hard to hit (and perhaps unreasonable - or unreasonable in relation to the data, meaning the data might be unreasonable in some fashion)
- it might be possible to investigate by looking at the target values you get using the 2016 growfactors and the starting weights (I think these would be the 2015 weights, but they might be 2011, I don't remember which the code uses - base year or prior year) and seeing if some are just way different from the 2016 targets - checking to see which variables are way off; the ones that are way off might have a bad 2016 growfactor or a bad 2016 target, or it may just be that the world changed in a way that is hard for a grown 2011 file to hit

**2029**
image

- iteration limit of 100 was hit
- objective is NaN -- not sure what to make of this, but note that the dual objective is huge
- targets were hit
- weight ratios were almost in range -- 0.54 to 1.47 -- didn't quite make +/- 0.45
- but notice the strange distribution -- almost all of them were driven down by ~0.45 or up by 0.45, so it is clear that the constraints can only be hit by jerking the weights around
- again I would investigate the constraints, looking for one (or more) of three possibilities: (1) bad (implausible) targets, (2) bad data resulting from bad growfactors, or (3) plausible targets and plausible growfactors, but the world (targets) changed in ways that our very old data has a hard time mimicking

**2030 (shown) and 2031**
image

- Things are really out of hand at this point and the constraints are not feasible.

I haven't looked at the targets but my intuition would be to look at the way the correction to the wage targets was implemented to see if the full set of new targets is internally inconsistent.

donboyd5:
Inconsistent targets could arise if, for example:

- You had a wage target for each agi range, PLUS a total wage target (which would not be a great idea, because it would be redundant), AND the sum of the wage targets by range was not consistent with the total wage target, or
- You had a wage target for each agi range, plus total targets for other income variables, plus a target for total income (e.g., agi), AND the sum of all the income targets was not consistent with the total income target

There are ways to be inconsistent, too, but these are obvious ones.

jdebacker:
An update on this branch.  I've tested this with the latest versions of Julia and Tulip and still get an iteration limit hit in 2029 and "infeasible" after that.

Will look into targets in more detail next, but was hoping a new solver would do the trick...

cc @andersonfrailey @donboyd5 @MattHJensen

PR #417, opened 2022-09-07
Add projection

bodiyang:
This PR will add projections including AGI, total incomes as well as detailed projections to the report


bodiyang:
[taxdata_report_2022-09-12.pdf](https://github.com/PSLmodels/taxdata/files/9551681/taxdata_report_2022-09-12.pdf) includes the detailed projections os AGI, total income and more items. 
[Validation_CBO_TaxCalc.xlsx](https://github.com/PSLmodels/taxdata/files/9551686/Validation_CBO_TaxCalc.xlsx) is the difference between CBO tax model and Tax-Calculator calculated projections

history/report.py
```diff
@@ -244,6 +244,27 @@ def report():
         template_args["puf_combined_table"] = None
         template_args["puf_income_table"] = None
         template_args["puf_payroll_table"] = None
+        template_args["puf_salaries_and_wages_table"] = None
```
jdebacker, line 247: This looks like it does what you need, but there's probably an easy way use few lines of code by, e.g., iterating over the dict and assigning `None` to all arguments and then just updating the value for the key `puf_msg`.

history/report_utils.py
```diff
@@ -526,13 +540,110 @@ def agg_liability_table(data, tax):
...
+
+    return final_df.to_markdown()
+
+
+def calculate_agi_share(calc, year):
+
+    calc.advance_to_year(year)
+    calc.calc_all()
+
+    vdf = calc.dataframe(['s006', 'c00100', 'expanded_income']).sort_values(by='expanded_income', ascending=False)
```
jdebacker, line 579: Line length looks quite long here.  You should probably wrap to 90 or less characters per line.

history/report_utils.py
```diff
@@ -589,7 +802,203 @@ def compare_calcs(base, new, name, template_args, plot_paths):
         aggs["Tax Liability"].append(new_aggs["combined"])
         aggs["Tax"].append("New Combined")
         aggs["Year"].append(year)
+
+        aggs2["Value"].append(cur_salary_wage)
```
jdebacker, line 806: Could put the variables and strings in lists and reduce the lines of code by using a loop or Python iterable method.  I think doing so would was the ability to follow what's happening and reduce errors.

bodiyang: have made the adjustment in this [ new commit](https://github.com/PSLmodels/taxdata/pull/417/commits/109cac61006c497417401e2d67cd31b922218cdc)

bodiyang:
@jdebacker As discussed in the tax-calculator meeting, I have completed the task to generate validation tables between CBO projections and tax-data/tax-calculator projections. Question now is how we want this validation results to look like --- 1. to include this in the tax-data update report; or 2. create a new folder "validation with CBO tax model" in tax-data as an independent item to include the new code and a new correlated report

jdebacker:
@bodiyang These changes look good to me and I think it's helpful to have these additional comparisons in the report.  

I would suggest, however, that in a future PR, work is done to highlight the key features in the report, which has grown quite lengthy.

@andersonfrailey I recommend merging this PR.

Issue #433, opened 2023-08-09
randomness of taxdata


bodiyang:
Taxdata have a behavior of randomness when producing puf.csv file. I did't find if there is a routine, but here are several cases.

1. This always happening: each run, the order of the columns are reshuffled, but the puf.csv is the same (the projection values are the same) ~ Taxdata is randomly reshuffled the order of columns. e.g. in one run e09900, g20500, ... MIDR; then in another run it produce MIDR, e09900, ....

2. The produced puf.csv are different, dramatically different, depending on whether I produce the file under taxdata-dev conda env or not. e.g. the wighted total iitax is $2205B when using puf produced under taxdata-dev; the iitax is $7452B when using puf produced not under this environm,ent 
~ The $2205B should be correct, basically match with CBO projections.  I understand that I should produce it under taxdata-dev env. However, I don't see why they cause such a big difference, since the role of taxdata-dev is to just claim which software packages are required ~ why causing difference in the calculation.

3. This happens sometimes ~ Taxdata are producing different result in each run, and as a consequence, different run's produced puf.csv will produce different projection values.   
Jason and I trying to produce puf file on our computers. Given that we are using same raw puf in 2011, same version of taxcalc and taxdata. The produced puf.csv looks different ~ e.g. showing different iitax weight total projection values
I feel this happening sometime, but not very sure ~ it would be a problem if this is real

 Any idea of why this is happening?
@andersonfrailey @jdebacker @MattHJensen

jdebacker:
I'm wondering whether the difference is with the `taxdata` package itself or something else (e.g., package versions for `taxdata` of `taxcalc` or their dependencies).

To get to the bottom of this, I think we need more systematic testing of this issue.

For example, I ran the `make all` command twice for `taxdata` (using the scripts in the `master` branch -- and in the `taxdata-dev` environment).  The resulting `puf.csv` files show no difference:

```
In [1]: import pandas as pd

In [2]: puf1 = pd.read_csv("./data/puf.csv")

In [3]: puf2 = pd.read_csv("./data/puf_20230814_309pm.csv")

In [4]: diff = puf1 - puf2

In [5]: diff.max()
Out[5]:
DSI         0
EIC         0
FLPDYR      0
MARS        0
MIDR        0
           ..
p22250      0
p23250      0
pencon_p    0
pencon_s    0
s006        0
Length: 94, dtype: int64

In [6]: diff.max().max()
Out[6]: 0
```

[...]

PR #435, opened 2023-09-13
Move impute_pencon targets to csv

andersonfrailey:
This PR moves the targets used in `impute_pencon.py` from a string in the python file to separate CSV files. The argument for doing this is that these targets are specific to the 2011, so when we update TaxData to be compatible with more recent versions of the PUF we'll need to use different targets. This PR will allow us to add new targets, while keeping the old ones for backwards compatibility.

I compared the resulting PUF to the PUF created from the master branch and while the column order is difference (see issue #433), the values in the PUF are the same.


andersonfrailey:
If there are no objections, I will merge this at the end of the day

jdebacker:
LGTM @andersonfrailey!

PR #436, opened 2023-09-13
Sort PUF columns

andersonfrailey:
This PR resolves issue #433. It simply sorts the PUF columns before they're exported after final prep. Using `createpuf.py` I was able to create two PUFs with identical MD5s.

cc @jdebacker @bodiyang


andersonfrailey:
If there are no objections, I will merge this at the end of the day.

bodiyang:
@andersonfrailey Good with merging, thanks Anderson

PR #449, opened 2025-06-04
Impute W-2 Wage Share variable for Qualified Business Income Deduction

bodiyang:
Imputation for W-2 wage share variable `PT_binc_w2_wages` to improve the accuracy of Qualified Business Income Deduction (QBID) amounts produced by Tax-Calculator when using `puf.csv` input data.   **It is important to note that the logic of QBID calculation in Tax-Calculator is correct.**  What makes QBID results inaccurate when using the current version of `puf.csv` is that the value of `PT_binc_w2_wages` is zero for every filing unit in the current version of the `puf.csv` file.   This PR is intended to fix this `puf.csv` data problem.


taxdata/puf/impute_qbid_w2.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Function that calculate the QBID aggregates through years.
+    """
+    qbided = []
+    df = data.copy()
+    df["PT_binc_w2_wages"] = ratio * df["qbinc"]
+    pol = tc.Policy()
+    recs = tc.Records(data=df)
+    calc0 = tc.Calculator(policy=pol, records=recs)
+    for year in range(2021, 2026):
```
jdebacker, line 61: @bodiyang Should the year range be an argument here? Any way to make it a function of QBID parameters in Tax-Calculator's current law baseline?  Or why not just go through the budget window used in Tax-Calculator?

bodiyang: > @bodiyang Should the year range be an argument here? Any way to make it a function of QBID parameters in Tax-Calculator's current law baseline? Or why not just go through the budget window used in Tax-Calculator?

Because I only wanted to select the year 2021 ~ 2025 QBID value as targets, not other years. It is noticed that there is a gap between the years before 2020 and the years after 2021 in CBO's published QBID aggregates. So the year 2021 and after probably better reflects the recent years QBID trajectory. Years before 2020 is a distraction for the model's optimization process.

jdebacker: Ok - understood. Thanks!

jdebacker (approved):

bodiyang:
> @bodiyang Your changes look good to me. Can you run `black -l 79 ./` from the top level `taxdata` directory (and then commit your modified tracked files) to fix the currently failing code formatting issue?

This PR is still not ready yet. There are some issues with incorporating the imputation script into the model.

 I'll notify when it's ready for review.

PR #453, opened 2025-09-02
Update taxdata-dev environment

bodiyang:
Tax-Data's conda environment `taxdata-dev` needs to be updated.

This PR will add the new packages reliances. 

Note: Since the new environment requires taxcalc >= 5.3.0, this PR can be merged after the completion of Tax-Data PR 452 and Tax-Calc version 5.3.0 release.


environment.yml
```diff
@@ -17,11 +17,15 @@ dependencies:
 - pulp
 - paramtools
 - pypandoc
+- openpyxl
```
jdebacker, line 20: @bodiyang, Are these 4 new dependencies used anywhere in the `taxdata` package?  e.g., I like `plotly`, but if `taxdata` doesn't call it, we don't want to include it as a dependency here.

bodiyang: @jdebacker it is required in generating report in history/report.py

environment.yml
```diff
@@ -17,11 +17,15 @@ dependencies:
...
+- lxml_html_clean
+- plotly
+- kaleido
 - selenium
 - python-chromedriver-binary
 - tabulate
 - pre-commit
-- taxcalc>=3.5.0
+- taxcalc>=5.3.0
```
jdebacker, line 28: I've noticed an issue trying to install TC 5.3 from conda.  Try to pip install instead (as done in current master branch).

environment.yml
```diff
@@ -17,13 +17,17 @@ dependencies:
 - pulp
 - paramtools
 - pypandoc
+- openpyxl
+- lxml_html_clean
+- plotly
+- kaleido
```
jdebacker, line 23: @bodiyang try: `python-kaleido`

